PR #4, opened 2016-11-10
sample and some basic memory checks

damienbod:
@ANurse @glennc


src/HealthCheck/HealthCheckBuilderExtensions.cs
```diff
@@ -21,6 +22,51 @@ public static HealthCheckBuilder AddUrlCheck(this HealthCheckBuilder builder, st
...
+                    return true;
+                }
+
+                return false;
+            });
+            
+            return builder;
+        }
+
+        public static HealthCheckBuilder AddWorkingSet64Check(this HealthCheckBuilder builder, long maxSize)
```
anurse: Let's just call these `AddWorkingSetCheck` and `AddPrivateMemorySizeCheck`. The `64` suffix on the properties is just an implementation detail.

PR #10, opened 2016-11-10
Damienbod/add cache busting

damienbod:
@anurse @glennc


src/HealthCheck/HealthCheckBuilderExtensions.cs
```diff
@@ -71,6 +71,8 @@ public static HealthCheckBuilder AddUrlCheck(this HealthCheckBuilder builder, st
         {
             builder.AddCheck($"UrlCheck ({url})", async () =>{
                 var httpClient = new HttpClient();
+                httpClient.DefaultRequestHeaders.Add("cache-control", "no-cache");
+                httpClient.DefaultRequestHeaders.Add("etag", DateTime.UtcNow.ToString());
```
anurse: ETags are only used in Response headers, so we don't need this line. `cache-control` is good though!

damienbod:
@anurse thanks updated

PR #14, opened 2016-11-12
Damienbod/add results healthcheck to response body

damienbod:
@anurse @glennc

### DIFF
--- a/src/HealthCheck/HealthCheckBuilderExtensions.cs
+++ b/src/HealthCheck/HealthCheckBuilderExtensions.cs
@@ -24,6 +24,7 @@ namespace HealthChecks
                 };
 
                 var httpClient = new HttpClient();
+                httpClient.DefaultRequestHeaders.Add("cache-control", "no-cache");
                 var response = await httpClient.GetAsync(url);
 
                 if (response.StatusCode == HttpStatusCode.OK)
@@ -110,24 +111,25 @@ namespace HealthChecks
             return builder;
         }
 
-        ////public static HealthCheckBuilder AddUrlCheck(this HealthCheckBuilder builder, string url, Func<HttpResponseMessage, bool> checkFunc)
-        ////{
-        ////    builder.AddCheck($"UrlCheck ({url})", async () =>{
-        ////        var healthCheckResult = new HealthCheckResult
-        ////        {
-        ////            Success = false,
-        ////            CheckStatus = CheckStatus.Failed,
-        ////            CheckType = "memory",
-        ////            Description = $"UrlCheck: {url}"
-        ////        };
+        public static HealthCheckBuilder AddUrlCheck(this HealthCheckBuilder builder, string url, Func<HttpResponseMessage, HealthCheckResult> checkFunc)
+        {
+            builder.AddCheck($"UrlCheck ({url})", async () =>
+            {
+                var healthCheckResult = new HealthCheckResult
+                {
+                    Success = false,
+                    CheckStatus = CheckStatus.Failed,
+                    CheckType = "memory",
+                    Description = $"UrlCheck: {url}"
+                };
 
-        ////        var httpClient = new HttpClient();
-        ////        httpClient.DefaultRequestHeaders.Add("cache-control", "no-cache");
-        ////        var response = await httpClient.GetAsync(url);
-        ////        return checkFunc(response);
-        ////    });
-        ////    return builder;
-        ////}
+                var httpClient = new HttpClient();
+                httpClient.DefaultRequestHeaders.Add("cache-control", "no-cache");
+                var response = await httpClient.GetAsync(url);
+                return checkFunc(response);
+            });
+            return builder;
+        }
 
         //TODO: Move this into a seperate project. Avoid DB dependencies in the main lib.
         //TODO: It is probably better if this is more generic, not SQL specific.
--- a/src/HealthCheck/HealthCheckBuilderExtensions.cs
+++ b/src/HealthCheck/HealthCheckBuilderExtensions.cs
@@ -17,10 +17,6 @@ namespace HealthChecks
         {
             var urls = urlItems.ToList();
             builder.AddCheck($"UrlChecks ({group})", async () => {
-                var healthCheckResult = new HealthCheckResult
-                {
-                    CheckStatus = CheckStatus.Unhealthy,
-                };
 
                 var successfulChecks = 0;
                 var description = new StringBuilder();
@@ -51,15 +47,14 @@ namespace HealthChecks
 
                 if (successfulChecks == urls.Count)
                 {
-                    healthCheckResult.CheckStatus = CheckStatus.Healthy;
+                    return HealthCheckResult.Healthy(description.ToString());
                 }
                 else if (successfulChecks > 0)
                 {
-                    healthCheckResult.CheckStatus = CheckStatus.Warning;
+                    return HealthCheckResult.Warning(description.ToString());
                 }
-                healthCheckResult.Description = description.ToString();
 
-                return healthCheckResult;
+                return HealthCheckResult.Unhealthy(description.ToString());
 
             });
             return builder;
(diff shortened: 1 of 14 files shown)


src/HealthCheck/HealthCheckResults.cs
```diff
@@ -0,0 +1,30 @@
...
+        public string AdditionalData { get; set; }
+
+        public string CheckType { get; set; }
+    }
+
+    public enum CheckStatus
+    {
+        Unknown,
+        Failed,
+        Ok,
```
anurse: `Healthy` would be a better term here, similar to above.

src/HealthCheck/HealthCheckResults.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+
+namespace HealthChecks
+{
+    public class HealthCheckResults
+    {
+        public List<HealthCheckResult> CheckResults { get; set; }
+    }
+
+    public class HealthCheckResult
```
anurse: Can we make this immutable (all properties read-only) with constructors or static factory methods (like `HealthCheckResult.Failed(string description)`)?

src/HealthCheck/HealthCheckResults.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+
+namespace HealthChecks
+{
+    public class HealthCheckResults
+    {
+        public List<HealthCheckResult> CheckResults { get; set; }
```
anurse: This should be read-only, initialized to a new empty list and of type `IList<HealthCheckResult>`.

src/HealthCheck/HealthCheckResults.cs
anurse: Also, we go pretty strict on the one-type-per-file rule. Even though they are small types, I think separate files would be best since I expect some of them to grow a bit (certainly `HealthCheckResults` and when we add XML docs, etc.)

src/HealthCheck/HealthCheckResults.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class HealthCheckResult
+    {
+        public bool Success { get; set; }
+        public CheckStatus CheckStatus { get; set; }
+
+        public string Description { get; set; }
+
+        public string AdditionalData { get; set; }
+
+        public string CheckType { get; set; }
```
anurse: We're going to need to think a bit about the design of this property I think. If the purpose is for consumers to read it and make decisions about the check result based on it, we'll need to make sure we're consistent with the naming.

I don't see it being used anywhere (except in rendering some JSON). Can we remove it for now? I'd rather wait until we have a good plan for this kind of property before putting something in, even in a prototype :)

src/HealthCheck/HealthCheckResults.cs
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public class HealthCheckResult
+    {
+        public bool Success { get; set; }
+        public CheckStatus CheckStatus { get; set; }
+
+        public string Description { get; set; }
+
+        public string AdditionalData { get; set; }
```
anurse: Same comment from `CheckType` below applies here. I'm not sure what the design for this is going to be yet, so I think we should remove it for now. Always easier to add stuff later :)

src/HealthCheck/HealthCheckResults.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace HealthChecks
+{
+    public class HealthCheckResults
+    {
+        public List<HealthCheckResult> CheckResults { get; set; }
+    }
+
+    public class HealthCheckResult
+    {
+        public bool Success { get; set; }
```
anurse: `CheckStatus` should give you this information already, right? I don't think we need two properties, especially if you could end up with `Success = true` and `CheckStatus = Unhealthy` or something weird like that.

src/HealthCheck/HealthCheckBuilderExtensions.cs
```diff
@@ -83,24 +190,37 @@ public static HealthCheckBuilder AddUrlCheck(this HealthCheckBuilder builder, st
...
+                    using (var connection = new SqlConnection(connectionString))
                     {
                         connection.Open();
                         using(var command = connection.CreateCommand())
                         {
                             command.CommandType = CommandType.Text;
                             command.CommandText = "SELECT 1";
                             var result = (int) await command.ExecuteScalarAsync();
-                            return result == 1;
+                            if(result == 1)
```
anurse: Give your files a quick auto-format in VS :). This should have a space between `if` and `(`

anurse:
Had some comments, and need some changes made, but generally looks good.

FYI: We haven't yet figured out where this will fit in the product, or if we have the engineering resources to take it on at the moment, so we probably won't have a lot of bandwidth to work on this in the short term. We can still keep the prototype maintained and do some small feature work, but probably won't want to go too far down the path of designing things until we actually have figured out how it will fit into the product :) /cc @glennc

src/HealthCheck/HealthCheckBuilderExtensions.cs
```diff
@@ -1,23 +1,94 @@
...
+        {
+            builder.AddCheck($"UrlChecks ({group})", async () => {
+                var healthCheckResult = new HealthCheckResult
+                {
+                    Success = false,
+                    CheckStatus = CheckStatus.Failed,
+                    CheckType = "url",
+                };
+
+                var successFulChecks = 0;
```
damienbod: ok

src/HealthCheck/HealthCheckBuilderExtensions.cs
damienbod: ok

src/HealthCheck/HealthCheckResults.cs
damienbod: ok

src/HealthCheck/HealthCheckResults.cs
damienbod: ok

src/HealthCheck/HealthCheckResults.cs
damienbod: ok

src/HealthCheck/HealthCheckResults.cs
damienbod: ok

src/HealthCheck/HealthCheckResults.cs
damienbod: ok

src/HealthCheck/HealthCheckResults.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public string AdditionalData { get; set; }
+
+        public string CheckType { get; set; }
+    }
+
+    public enum CheckStatus
+    {
+        Unknown,
+        Failed,
```
damienbod: ok

src/HealthCheck/HealthCheckResults.cs
damienbod: ok

damienbod:
@anurse Thanks for your feedback. I've implemented the review findings.

Greetings Damien

src/HealthCheck/HealthCheckResult.cs
```diff
@@ -0,0 +1,55 @@
+﻿namespace HealthChecks
+{
+    public class HealthCheckResult
+    {
+        private CheckStatus _checkStatus;
+
+        private string _description;
+
+        public CheckStatus CheckStatus
```
anurse: In C# 6 you can actually write these as just getter-only properties, which is generally what we prefer. You'll still be able to set them in a constructor, despite not having a defined setter, but **only** in a constructor. You can add a `private` constructor and still have the `public static` factory methods that you have below (which I like!).

So something like this:

```csharp
public CheckStatus CheckStatus { get; }
public string Description { get; }

private HealthCheckResult(CheckStatus checkStatus, string description)
{
    CheckStatus = checkStatus;
    Description = description;
}
```

src/HealthCheck/HealthCheckResults.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Collections.Generic;
+
+namespace HealthChecks
+{
+    public class HealthCheckResults
+    {
+        private IList<HealthCheckResult> _checkResults { get; set; }
```
anurse: You actually have two properties here, we can just do it with one getter-only property (as above). And, you don't even need a constructor, you can initialize it in the property definition itself:

```csharp
public class HealthCheckResults
{
    public IList<HealthCheckResult> CheckResults { get; } = new List<HealthCheckResult>();
}
```

src/HealthCheck/HealthCheckResult.cs
damienbod: ok

src/HealthCheck/HealthCheckResults.cs
damienbod: ok

damienbod:
@anurse thanks, I fixed the code-style things

Greetings Damien

anurse:
Awesome! Thanks for the updates!

damienbod:
cool, thanks

PR #23, opened 2018-02-27
Update Microsoft.Net.Http.Headers.cs

joseftw:
Adds the following security headers:

- Referrer-Policy
- Strict-Transport-Security
- X-Content-Type-Options
- X-Frame-Options
- X-Xss-Protection


joseftw:
Something weird happend with the commit messages, will fix it and send create a new one.

PR #24, opened 2018-02-27
Adds some security headers

joseftw:
Adds the following headers:

- Referrer-Policy
- Strict-Transport-Security
- X-Content-Type-Options
- X-Frame-Options
- X-XSS-Protection


joseftw:
This is most likely the wrong repo...

PR #25, opened 2018-06-13
Create a prototype of Microsoft.AspNetCore.App.RefOnly

natemcmaster:
Prototype what it would look like for aspnetcore to follow the pattern set by Microsoft.NETCore.App. Differences between Microsoft.AspNetCore.App.RefOnly and Microsoft.AspNetCore.App

* The package only contains reference assemblies in ref/netcoreapp2.1/
* The package does not have dependencies
* There are not implementation binaries in lib/

@mikeharder do you have tools to do perf testing? @davidfowl  want to know the impact of having few packages in the restore graph.

cc @muratg


davidfowl:
Awesome @natemcmaster. Lets sit down tomorrow and try some things.

PR #26, opened 2018-06-15
Produce an LZMA with the refonly package

natemcmaster:
Comparison:

## Old LZMA
See https://github.com/aspnet/Universe/issues/1124 for more details on the composition of the current NuGetFallbackFolder
* 1,010 MB once expanded on disk
  * ~670 MB of xml docs
  * 181 MB of binaries
## LZMA with Microsoft.AspNetCore.App.RefOnly

* 113 MB once expanded on disk
  * 49 MB of xml docs
  * 41MB of binaries

*And it can still get smaller!* This prototype still bundles the implementation binaries in Microsoft.AspNetCore.App.RefOnly. We could strip these down to actual ref assemblies.

@DamianEdwards @Petermarcu @davidfowl @weshaggard

### DIFF
--- a/pkg/.gitignore
+++ b/pkg/.gitignore
@@ -1,5 +1,3 @@
-feed/
-publish/
 obj/
 ref/
 .dotnet/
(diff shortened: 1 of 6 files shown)


davidfowl:
Nice! Merge it! This is great data! The improvements looks nice.

PR #28, opened 2018-10-09
Merge aspnet/Proxy into this repository

natemcmaster:
Part of https://github.com/aspnet/AspNetCore-Internal/issues/846. Condensing repos.

This merges the source code of aspnet/Proxy into this repository, along with all of its git history. You can still run "git blame" and "git log" on files to see the history of the .cs files.


natemcmaster:
FYI since this is an experiment, git history may not be as interesting. We could also squash history as a part of the move. Thoughts?

Eilon:
Eh I'd say keep the history. Why not.

natemcmaster:
Ok, sounds good.

PR #30, opened 2018-10-12
import dotnet-monitor code

anurse:
This is a simple import of the code from prototyping work done in https://github.com/anurse/dotnet-monitor. I plan to do some clean up after merging including removing the "event pipe protocol" and in-proc diagnostics server. I want to check them in during the initial import so we have the code in source control, but I think we're moving away from that pattern with the 2.2 work on file-based sharing. The protocol is a hack anyway.

I can start looking in to things like collecting traces from the file-based config (assuming that's in the latest 2.2 builds) and collecting dumps next.

Apparently AspLabs is the preference over a separate repo, so we'll pack it in here.

cc @natemcmaster @Eilon @muratg FYI regarding this new project in AspLabs. `dotnet-monitor` is an umbrella term (for now) for .NET diagnostics improvements across the stack being considered in 3.0. This is bringing in some of the initial prototype work.

No CI needs here yet, but we'll probably pull something together in DevOps in the coming weeks as 2.2 work winds down. For the most part I can run with the same stuff I've learned in SignalR "Classic"


natemcmaster:
We can use #31 to track work to add CI.

Eilon:
This is the thing that demonstrated that the thing we did in 2.2 actually does good things? If so, let's get this thing checked in!

PR #34, opened 2018-10-12
dotnet-dump for windows/linux, full dumps only

anurse:
Early prototype of `dotnet-dump`. Supports Windows only and just captures a dump of all memory.

Based on anurse/add-dotnet-monitor

### DIFF
--- a/src/dotnet-monitor/Directory.Build.props
+++ b/src/dotnet-monitor/Directory.Build.props
@@ -1,6 +1,5 @@
 <Project>
     <PropertyGroup>
         <LangVersion>latest</LangVersion>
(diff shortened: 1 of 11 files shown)


anurse:
There are some non-sequitur changes because the initial import has some build failures in it.

anurse:
Linux should work now, via locating `createdump` next to `libcoreclr.so`. Couldn't test it fully since `createdump` doesn't work on WSL :(

anurse:
Waiting until #30 is merged before I merge this, since it's based on top of it. I'll probably just merge it on Monday if nobody has complained ;P

davidfowl:
OK I'll play with this branch

PR #35, opened 2018-10-16
add dotnet-collect command that uses EventPipe

### LINKED ISSUE #32: Change dotnet-trace to collect via File-based eventpipe config.
https://dotnetfeed.blob.core.windows.net/orchestrated-release-2-2/20181010-04/final/assets/Sdk/2.2.100-preview3-009428/dotnet-sdk-2.2.100-preview3-009428-win-x64.zip has support for File-based eventpipe configuration. It does not currently have support for automatic flushing, but we could have a "start/stop" tracing mechanic.

anurse:
```
Collects Event Traces from .NET processes

Usage: dotnet-collect [options]

Options:
  -c|--config-path <CONFIG_PATH>  The path of the EventPipe config file to write, must be named [AppName].eventpipeconfig and be in the base directory for a managed app.
  -p|--process-id <PROCESS_ID>    Filter to only the process with the specified process ID.
  -o|--output <OUTPUT_DIRECTORY>  The directory to write the trace to. Defaults to the current working directory.
  --buffer <BUFFER_SIZE_IN_MB>    The size of the in-memory circular buffer in megabytes.
  --provider <PROVIDER_SPEC>      An EventPipe provider to enable. A string in the form '<provider name>:<keywords>:<level>'. Can be specified multiple times to enable multiple providers.
  -?|-h|--help                    Show help information
```

The `-c|--config-path` is the only required option (though no providers will be enabled ;)). As an example, to collect all events from the .NET Runtime event source:

```
dotnet collect -c ./path/to/AppName.eventpipeconfig --provider Microsoft-Windows-DotNETRuntime
```

Files will be written to the output directory (`-o|--output`, defaults to current working directory) in the form `[appname].[processId].netperf` and can be viewed in PerfView.

Fixes #32

### DIFF
--- a/src/dotnet-monitor/dotnet-monitor.sln
+++ b/src/dotnet-monitor/dotnet-monitor.sln
@@ -20,6 +20,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "dotnet-analyze", "src\dotne
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "dotnet-dump", "src\dotnet-dump\dotnet-dump.csproj", "{933327B7-87FC-4D1B-8AF4-C6A07DE30AA4}"
 EndProject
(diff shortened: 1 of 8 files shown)


anurse:
FYI https://github.com/dotnet/coreclr/issues/20454 is tracking an issue where the eventpipeconfig file has to be next to `dotnet.exe` rather than next to the managed entry point dll.

anurse:
I'llformatyou

PR #66, opened 2018-11-19
Set the MINIDUMP_TYPE enum values correctly

stuarthillary:
Decorate the MINIDUMP_TYPE enum with the Flags attribute
Set the values of the MINIDUMP_TYPE enum to those found in minidumpapiset.h file from the Windows 10 SDK


anurse:
Nice :). Sounds good to me.

PR #74, opened 2018-11-27
Use Arcade for local command-line builds

anurse:
This PR adds .NET Arcade to this repo and a root build script to launch the restore, build, test and pack phases of the build.

This also includes an `azure-pipelines.yml` file that should be able to build the project in AzDO, but I'm going to work on that in a separate branch and send a second PR if other changes are required. It's the same one I used in Microsoft/clrmd though so it *should* just work (famous last words).

Anything under `eng/common` should be ignored since it's part of Arcade's shared toolset and will just be replaced as new versions come in. There's one exception in `eng/common/tools.ps1` that I flagged with a comment.

TODO:
* [x] Add appropriate NuGet feeds. It was building on my machine because I had packages synced. Looks simple enough.


.gitignore
```diff
@@ -1,29 +1,132 @@
-[Oo]bj/
-[Bb]in/
-TestResults/
-.nuget/
-*.sln.ide/
-_ReSharper.*/
-packages/
-shared/
+syntax: glob
```
natemcmaster: Just curious...what's this?

src/WebHooks/src/Microsoft.AspNetCore.WebHooks.Receivers.AzureContainerRegistry/Microsoft.AspNetCore.WebHooks.Receivers.AzureContainerRegistry.csproj
```diff
@@ -1,11 +1,10 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
```
natemcmaster: ?

src/Proxy/Proxy.sln
```diff
@@ -3,11 +3,17 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.26124.0
 MinimumVisualStudioVersion = 15.0.26124.0
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.AspNetCore.Proxy", "src\Microsoft.AspNetCore.Proxy.csproj", "{01C51C11-BDCD-4D69-A829-2BA3846D431C}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Proxy", "src\Microsoft.AspNetCore.Proxy\Microsoft.AspNetCore.Proxy.csproj", "{01C51C11-BDCD-4D69-A829-2BA3846D431C}"
```
natemcmaster: The layout was intentional and meant to match the patterns we're using in aspnet/AspNetCore. Why are you reorganizing the code? Is it something Arcade requires?

src/HttpRepl/Microsoft.HttpRepl/Microsoft.HttpRepl.csproj
```diff
@@ -3,11 +3,13 @@
...
     <TargetFramework>netcoreapp2.1</TargetFramework>
-    <PackAsTool>true</PackAsTool>
     <AssemblyName>dotnet-httprepl</AssemblyName>
-    <LangVersion>latest</LangVersion>
     <Description>Command line tool to for making HTTP calls and viewing their results.</Description>
     <PackageTags>dotnet;http;httprepl</PackageTags>
+
+    <IsPackable>true</IsPackable>
+    <PackAsTool>true</PackAsTool>
+    <ToolCommandName>dotnet-httprepl</ToolCommandName>
```
natemcmaster: Unnecessary, right? Should be inferred from AssemblyName.

natemcmaster:
Also, any reason we shouldn't enable PR validation using azure-pipelines.yml?

.gitignore
anurse: I grabbed the `.gitignore` from the Arcade template and pulled it in. The `syntax: glob` is just a way of "forcing" git to only use the glob syntax. Though I can't find docs for that.. I'll yank that line. I do think we should align to more consistent gitignores though (this one embeds the "official" [VisualStudio.gitignore](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore) file as well)

eng/common/tools.ps1
```diff
@@ -0,0 +1,284 @@
...
+
+  # Source Build uses DotNetCoreSdkDir variable
+  if ($env:DotNetCoreSdkDir -ne $null) {
+    $env:DOTNET_INSTALL_DIR = $env:DotNetCoreSdkDir
+  }
+
+  # Find the first path on %PATH% that contains the dotnet.exe
+  if ($useInstalledDotNetCli -and ($env:DOTNET_INSTALL_DIR -eq $null)) {
+    # PATCH: Fix https://github.com/dotnet/arcade/issues/1251
+    $env:DOTNET_INSTALL_DIR = ${env:PATH}.Split(';') | where { ($_ -ne "") -and (Test-Path (Join-Path $_ "dotnet.exe")) } | select -first 1
```
anurse: This is the patch I had to make to work around an issue on my machine (and any other machine with multiple `dotnet.exe` instances on the path). Without this fix, if you have multiple instances of `dotnet.exe`, the script will *always* install a local copy to the `.dotnet` folder **in this repo** instead of using your local one, even if you have the right version.

anurse:
> Also, any reason we shouldn't enable PR validation using azure-pipelines.yml?

@natemcmaster Nope! I just wanted to get this part in and start working on the rest. The YAML file supports both scenarios. I'm looking at the AzDO stuff now.

src/WebHooks/src/Microsoft.AspNetCore.WebHooks.Receivers.AzureContainerRegistry/Microsoft.AspNetCore.WebHooks.Receivers.AzureContainerRegistry.csproj
anurse: 🤷‍♂️

I'll remove it and see if anything breaks :)

src/Proxy/Proxy.sln
anurse: I'll revert it, it's not required by Arcade. I changed it because it looked like someone messed things up since it's so different from other .NET projects (corefx-exempted because they're weired and special) and it's inconsistent from how a multi-project folder would look.

anurse:
Got the PR build running, as well as the ["official" build](https://dev.azure.com/dnceng/internal/_build/index?context=allDefinitions&path=%5Caspnet%5CAspLabs&definitionId=249&_a=completed) (insofar as there is an official build of the labs stuff ;))

PR #79, opened 2018-12-08
Turn off input canonicalization (icanon) to avoid the password cursor showing up on mac

### LINKED ISSUE #67: dotnet-httprepl - Lock icon shows up when backspace or arrows are pressed on MacOS
When users use the backspace keystroke when in the tool to backspace over previously-written content, the icon changes to the "lock" icon, which to a MacOS user is usually a cue that they need to enter their password. The gif below shows the issue in practice. 

mac-typing-issue mp4

This is a disruptive experience.

mlorbetske:
Fixes #67


mlorbetske:
cc @vijayrkn

PR #80, opened 2018-12-08
Add cycle detection for JSON pointer resolution

### LINKED ISSUE #69: dotnet-httprepl swagger stackoverflowexception
While I am setting the httprepl swagger using the command :
`set swagger http://example.com/api/v2/swagger.json` 
I get the response : 

Process is terminating due to StackOverflowException.

The  swagger.json file size is 262 KB.

### LINKED ISSUE #70: StackOverflowException upon 404 for initial HEAD request
If the API does not accept a HEAD request for the "set base" URL and returns a 404 then the REPL errors out with a StackOverflowException.  ("Process is terminating due to StackOverflowException.")

The base URL should probably accept a HEAD request? But, in case it does not, like the one I was trying, may be a more meaningful error will be helpful. 

Even though the base URL returned a 404, REPL still makes a request to a default swagger URL and in my case, that did return 200. So may be if both return 404 only then return an error?

Was able to observe this behavior with Fiddler as follows:

image

image

λ dotnet --version
2.2.100-preview2-009404

Also, Some of the keywords for HELP seem to be missing when I tried it out. I tried it in regular CMD prompt as well and got the same result. 

image

image

mlorbetske:
Write the resolved token into arrays when resolving their children

Fixes #69 
Fixes #70

### DIFF
--- a/src/HttpRepl/Microsoft.HttpRepl/OpenApi/PointerUtil.cs
+++ b/src/HttpRepl/Microsoft.HttpRepl/OpenApi/PointerUtil.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Net.Http;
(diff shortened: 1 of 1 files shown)


mlorbetske:
cc @vijayrkn

PR #84, opened 2018-12-13
Updating the pre-release version for httpRepl to preview

vijayrkn:


vijayrkn:
This did not work. Based on the documentation here it does produce dev & ci respectively but is still picking up the PreReleaseVersionLabel from Version.props. I will check why.

https://github.com/dotnet/arcade/blob/1a192b6f23654fc0bda250552800a40fd56c6d65/Documentation/CorePackages/Versioning.md#package-version-generation

vijayrkn:
Ah. it is because it gets reset to a different variable. I will apply this private property as a work around.

```
    <_PreReleaseLabel>$(PreReleaseVersionLabel)</_PreReleaseLabel>
    <_PreReleaseLabel Condition="'$(ContinuousIntegrationBuild)' == 'true' and '$(OfficialBuild)' != 'true'">ci</_PreReleaseLabel>
    <_PreReleaseLabel Condition="'$(ContinuousIntegrationBuild)' != 'true'">dev</_PreReleaseLabel>
```

@joeloff - fyi - in case you run into this with arcade.

PR #169, opened 2019-02-15
Move projects to target netcoreapp3.0

mlorbetske:


vijayrkn:
@bradygaster - you want this change as well?

bradygaster:
If this would force customers to install .NET Core 3.0 preview to be able to use the http-repl, then please don't push that to public nuget. we want LTS & Current customers to be able to use it without upping to 3.0. *for now*

vijayrkn:
I have reverted this change for now. The latest package does not have this change. Will talk to Mike before getting this change back in.

PR #176, opened 2019-02-26
Add ComponentsElectron experimental projects

SteveSandersonMS:
As requested a while ago, this PR is to migrate the BlazorElectron experiment from https://github.com/SteveSandersonMS/BlazorElectronExperiment.Packages and https://github.com/SteveSandersonMS/BlazorElectronExperiment.Sample to this asplabs repo.

As part of this process, I've also updated it to run on the latest ASP.NET Core Components packages, and generally renamed everything to be consistent with it being a "Components" thing not a "Blazor" thing (since there's no WebAssembly in this model).

~The repo is configured *not* to build these new packages in CI, because they are `netcoreapp3.0`, and this repo only supports 2.x. When the repo is later updated to support 3.0 (@natemcmaster, is there a plan?), we can trivially enable building the ComponentsElectron solution. It's still valuable to have the sources here even if they only build when you clone and build locally.~ Update: Now it does build on 3.0 in CI.

The overall goal with this PR is to move the experiment to a team-owned location rather than a personal GitHub account, so that others can better make sense of how to proceed with prototyping. It's explicitly a non-goal to make the runtime packages production grade, as this will involve a larger investment of relayering and cleanup effort, some of which will require changes in the real Components.* packages. We'll only undertake that larger effort if proven to be justified.


natemcmaster:
> When the repo is later updated to support 3.0

Can you do that in this PR? Should be as simple as updating the tools.dotnet property in global.json

SteveSandersonMS:
@natemcmaster Unfortunately it's not that simple. When any `netcoreapp3.0` project builds in this repo's CI setup, it fails with:

```
NU1102: Unable to find package Microsoft.WindowsDesktop.App with version (>= 3.0.0-preview3-27421-1) [/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/ComponentsElectron.sln]
/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/testassets/TestApp/TestApp.csproj : error NU1102:   - Found 1 version(s) in nuget.org [ Nearest version: 3.0.0-alpha-27128-4 ] [/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/ComponentsElectron.sln]
/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/testassets/TestApp/TestApp.csproj : error NU1102:   - Found 1 version(s) in https://dotnet.myget.org/F/dotnet-core/api/v3/index.json [ Nearest version: 3.0.0-alpha-27128-4 ] [/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/ComponentsElectron.sln]
/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/testassets/TestApp/TestApp.csproj : error NU1102:   - Found 0 version(s) in https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json [/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/ComponentsElectron.sln]
/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/testassets/TestApp/TestApp.csproj : error NU1102:   - Found 0 version(s) in https://dotnet.myget.org/F/sourcelink/api/v3/index.json [/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/ComponentsElectron.sln]
/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/testassets/TestApp/TestApp.csproj : error NU1102:   - Found 0 version(s) in https://www.myget.org/F/anurse-clrmd/api/v3/index.json [/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/ComponentsElectron.sln]
/Users/vsts/agent/2.147.1/work/1/s/src/ComponentsElectron/testassets/TestApp/TestApp.csproj : error NU1102:   - Found 0 version(s) in https://www.myget.org/F/xunit/api/v3/index.json [/Users/vsts/agent/2.147.1/work/1/s/s
```

This builds file locally when using the same .NET SDK. I *think* it's supposed to get the `Microsoft.WindowsDesktop.App` that's bundled with the SDK, so don't know why it can't in CI. Do you know why?

If there's no straightforwards resolution, I'd rather *not* deal with the .NET Core 3.0 upgrade in this PR, as it doesn't block anything to have it not build these specific packages in CI in the short term. The goal for this PR is just to move the sources, not to set up a long-term build and distribution mechanism.

SteveSandersonMS:
Oh, I can repro it locally now, so can probably fix this.

SteveSandersonMS:
OK, it's a mess. I can make the new `netcoreapp3.0` projects build but only by copying in a couple of the `Workarounds.targets` entries from the `aspnetcore` repo to deal with some SDK issues. But then some of the existing 2.0-targeting projects get broken. It's *not* as simple as changing `global.json` 😢 

So, I *don't* wish to consider upgrading this repo to .NET Core 3.0 as in scope for this PR. Fortunately it doesn't block anything to have it as-is. OK, @natemcmaster?

natemcmaster:
There are no plans currently to update this repo to use the 3.0. No reason
it shouldn't be updated...just no one is available to do the work right
now. If this is something your code needs to be usable in this repo, I'd
suggest spending some more time on this. Code that doesn't build isn't
really useful. The SDK team may be able to help you identify why the 3.0
SDK seems to be broken.

cc @Eilon

SteveSandersonMS:
It does build locally, which is the useful part. Having it build in CI would be nice but is not something we require for any particular reason right now. This Electron feature is very much an unsupported experiment, so I can’t justify further yak-shaving.

@Eilon Are you happy with this being here? Alternatively we could create a separate repo for it.

SteveSandersonMS:
Right, maybe I shouldn't have continued the yak-shaving, but I did, and eventually got the whole repo to agree to be built with the 3.0 SDK in CI. This should be ready now. 😅

src/ComponentsElectron/src/Microsoft.AspNetCore.Components.Electron/ElectronRenderer.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        private void CaptureAsyncExceptions(Task task)
+        {
+            task.ContinueWith(t =>
+            {
+                if (t.IsFaulted)
+                {
+                    UnhandledException?.Invoke(this, t.Exception);
+                }
+            });
```
rynowak: Should this use `TaskScheduler.Current`? (or just `await`).

This looks like it copied an old bug.

SteveSandersonMS: Good point - fixed.

SteveSandersonMS:
Merging because the build-related stuff is also resolved, plus we got agreement to move this content here by email earlier.

Eilon:
> @Eilon Are you happy with this being here? Alternatively we could create a separate repo for it. Or we could argue that this is a “sample” (some of it is) so should go in that repo, assuming it doesn’t use the same build system.

Yes totally fine. Sorry it took me a while to get to this.

PR #188, opened 2019-05-04
Make Blazectron work

pranavkm:

PR #189, opened 2019-05-05
Switch to non `-Int` pools and clear NuGet caches

dougbu:
- aspnet/AspNetCore-Internal#2436
- aspnet/AspNetCore#9767

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,10 +44,10 @@ jobs:
     - job: Windows
       pool:
         ${{ if eq(variables['System.TeamProject'], 'public') }}:
-          name: NetCorePublic-Int-Pool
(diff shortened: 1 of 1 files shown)


dougbu:
@javiercn @pranavkm from the errors in https://dev.azure.com/dnceng/public/_build/results?buildId=179785, it appears AspLabs uses a Microsoft.AspNetCore.Components.Server namespace that no longer exists. Suspect a package was cached containing that namespace. What's the fix?

dougbu:
@pranavkm #188 failed all checks but was merged. See more comments in that PR and fix 'master' in this repo.

Checking this in because the build failures are consistent with 'master'. The problem is _not_ clearing the NuGet caches for example.

PR #191, opened 2019-05-10
dotnet-iis global tool?

anurse:
Is it cool? Is it crazy? YOU be the judge!

This is a **suuuper** hacky first pass at an idea. I'm opening a PR so it can go in to master and people can play with it. There are many problems with this :).


anurse:
The build has a sad and while it is not required I want to make it happy before merging. Taking a quick look.

anurse:
Looks like the ComponentsElectron experiment is just very out of date. I'm disabling it in the build since I don't think it's a priority (fyi @pranavkm)

PR #192, opened 2019-06-10
Update ComponentsElectron sample to preview 6

SteveSandersonMS:


SteveSandersonMS:
I'm not sure why it still isn't restoring the packages correctly in CI, but I'll come back to this when I can. For now this is enough to be able to build and run locally.

Maybe once preview 6 ships this will get much simpler as it will get the packages from the standard public feed.

pranavkm:
Closing this since this out of date now.

PR #194, opened 2019-06-23
Move to 3.0 Preview 6 SDK

dougbu:
@anurse any reason not to do this?

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.0.100-preview-010184"
+    "dotnet": "3.0.100-preview6-012264"
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19121.5"
--- a/global.json
+++ b/global.json
@@ -1,4 +1,7 @@
 {
+  "sdk": {
+    "version": "3.0.100-preview6-012264"
+  },
   "tools": {
     "dotnet": "3.0.100-preview6-012264"
   },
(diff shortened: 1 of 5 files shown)


src/WebHooks/samples/Directory.Build.props
```diff
@@ -7,5 +7,6 @@
     <StandardTestWebsiteTfms>$(DeveloperBuildTestWebsiteTfms)</StandardTestWebsiteTfms>
     <StandardTestWebsiteTfms Condition=" '$(DeveloperBuild)' != 'true' ">netcoreapp2.1</StandardTestWebsiteTfms>
     <StandardTestWebsiteTfms Condition=" '$(DeveloperBuild)' != 'true' AND '$(OS)' == 'Windows_NT' ">$(StandardTestWebsiteTfms);net461</StandardTestWebsiteTfms>
+    <IsPackable>true</IsPackable>
```
JunTaoLuo: We don't usually set samples to be packable: https://github.com/aspnet/AspNetCore/blob/4f67ff921673d7367c44ea1840c7a731628bf9bf/Directory.Build.targets#L49

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.0.100-preview-010184"
+    "dotnet": "3.0.100-preview6-012264"
```
JunTaoLuo: Why don't we add a requirement for the same SDK version like we do in every other repo?
```json
"sdk": {
    "version": "3.0.100-preview6-012264"
  },
```

anurse:
Always Be Updating ;)

src/WebHooks/samples/Directory.Build.props
dougbu: This is an Web SDK annoyance. It sets `$(WarnOnPackingNonPackableProject)` to `true` for all web projects. I'll try adding `<WarnOnPackingNonPackableProject>false</WarnOnPackingNonPackableProject>` instead.

dougbu:
:up::date: to fix build

PR #223, opened 2019-09-19
Update ComponentsElectron to 3.0.0 RC1

SteveSandersonMS:
We can do a further update to 3.0.0 final afterwards, and it will be completely trivial since there are no API changes. We just have to update the package version it depends on.

A large proportion of the diff here is updates to the shared sources (particularly, all of the `.ts` sources).


SteveSandersonMS:
Note: the CI build **fails** currently because this repo is still trying to build using the 3.0.0-preview6 SDK for some reason. We need to update the repo before we can merge this PR.

GoranHalvarsson:
Wohoo, nice @SteveSandersonMS. I'm glad it's still alive :)

src/ComponentsElectron/src/Microsoft.AspNetCore.Components.Electron/NullDispatcher.cs
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Components.Electron
```
pranavkm: Would it be incorrect to use the default dispatcher?

danroth27:
@pranavkm @dougbu Should this PR be unblocked at this point?

dougbu:
Perhaps but it needs to be rebased on latest 'master' or perhaps 'release/3.1'. Build shouldn't fail as badly after that.

pranavkm:
These changes were merged as part of https://github.com/aspnet/AspLabs/pull/224

PR #244, opened 2019-12-23
Attempting to fix MacOSX build break.

NTaylorMullen:
- Given that I can't reproduce this issue locally this PR is mostly an attempt to get Azure DevOps to run a build on my behalf to see if something as simple as moving to yarn results in the build being fixed.


NTaylorMullen:
This looks to have fixed the issue, i'll incorporate it into the latest merge PR

PR #245, opened 2020-01-08
Add gRPC HTTP API

JamesNK:
Read me preview: https://github.com/aspnet/AspLabs/blob/jamesnk/grpc-httpapi/src/GrpcHttpApi/README.md


src/GrpcHttpApi/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+HTTP+JSON APIs for your gRPC services. No duplication!
+
+![gRPC loves REST](grpc-rest-logo.png "gRPC loves REST")
+
+### Usage:
+
+1. Add a package reference to `Microsoft.AspNetCore.Grpc.HttpApi`.
+2. Register services in *Startup.cs* with `AddGrpcHttpApi()`.
+2. Include `<Protobuf>` reference to *google/api/http.proto* and *google/api/http.proto*.
```
Pilchie: One of these should have been `annotations.proto`?

JamesNK: Yes, good catch.

PR #290, opened 2020-06-24
Use async using to get none IO blocking stream.

Anduin2017:


davidfowl:
Thanks

Anduin2017:
Seems the code can't pass building after merging this. So I opened another pull request which fixes that. #291

PR #293, opened 2020-07-17
Fix the build by removing async using

Tratcher:
Reverts https://github.com/aspnet/AspLabs/pull/290 to fix the build.

It looks like IAsyncDisposable isn't available in netstandard2.0.


Legends:
> Reverts #290 to fix the build.
> 
> It looks like IAsyncDisposable isn't available in netstandard2.0.

Is there a way/tool that we can use to check availability?

Tratcher:
> > Reverts #290 to fix the build.
> > It looks like IAsyncDisposable isn't available in netstandard2.0.
> 
> Is there a way/tool that we can use to check availability?

https://apisof.net/catalog/System.IAsyncDisposable

PR #294, opened 2020-07-17
Add a ProxyProtocol sample

Tratcher:
This is a sample for parsing the Proxy Protocol as a Kestrel connection middleware. The protocol allows level 4 proxies to forward information as a connection preface. The preface needs to be removed before HTTP parsing can happen.

We're publishing this to labs because multiple parties have expressed interest in it. We know it's not exhaustive, it (e.g. it only handles IPv4 addresses), but it's a good starting point for interested parties.

See https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt Section 2.2

Azure private link is one such proxy that sends this data:
https://docs.microsoft.com/en-us/azure/private-link/private-link-service-overview#properties (EnableProxyProtocol)

Ngnix has a first class implementation of this spec (inbound). https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/#configuring-nginx-to-accept-the-proxy-protocol

### DIFF
--- a/src/ProxyProtocol/Client.Sample/Program.cs
+++ b/src/ProxyProtocol/Client.Sample/Program.cs
@@ -1,0 +1,95 @@
+using System;
+using System.IO;
+using System.Net.Security;
+using System.Net.Sockets;
+using System.Reflection.Metadata;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Client.Sample
+{
+    class Program
+    {
+        static async Task Main(string[] args)
+        {
+            var serverAddress = new Uri("https://localhost:5001");
+
+            Console.WriteLine("Ready");
+            Console.ReadKey();
+
+            await ConnectAsync(serverAddress);
+        }
+
+        private static async Task ConnectAsync(Uri serverAddress)
+        {
+            using var client = new TcpClient();
+            Console.WriteLine($"Connecting to {serverAddress}");
+            await client.ConnectAsync(serverAddress.Host, serverAddress.Port);
+
+            Stream stream = client.GetStream();
+
+            if (serverAddress.Scheme == Uri.UriSchemeHttps)
+            {
+                Console.WriteLine("Negotiating TLS");
+                var sslStream = new SslStream(stream);
+                await sslStream.AuthenticateAsClientAsync(new SslClientAuthenticationOptions()
+                {
+                    TargetHost = serverAddress.Host,
+                    RemoteCertificateValidationCallback = (_, __, ___, ____) => true,
+                });
+
+                stream = sslStream;
+            }
+
+            Console.WriteLine("Sending message");
+            var data = CreateMessage(serverAddress);
+            await stream.WriteAsync(data);
+
+            Console.WriteLine("Reading response");
+            using var reader = new StreamReader(stream);
+            var response = await reader.ReadToEndAsync();
+
+            Console.WriteLine(response);
+
+            stream.Dispose();
+        }
+
+        /// Proxy Protocol v2: https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt Section 2.2
+        private static byte[] CreateMessage(Uri serverAddress)
+        {
+            var memoryStream = new MemoryStream();
+            var prefix = new byte[] {
+                // Preamble(12 bytes) : 0D-0A-0D-0A-00-0D-0A-51-55-49-54-0A
+                0x0D, 0x0A, 0x0D, 0x0A, 0x00, 0x0D, 0x0A, 0x51, 0x55, 0x49, 0x54, 0x0A,
+                //  -21 Version + stream
+                0x21,
+                //  -11 TCP over IPv4
+                0x11,
+                //  -00-14 length
+                0x00, 0x14,
+                //  -AC-1C-00-04 src address
+                0xAC, 0x1C, 0x00, 0x04,
+                //  -01-02-03-04 dest address
+                0x01, 0x02, 0x03, 0x04,
+                //  -D7-9A src port
+                0xD7, 0x9A,
+                //  -13-88 dest port
+                0x13, 0x88,
+                //  -EE PP2_TYPE_AZURE
+                0xEE,
+                //  -00-05 length
+                0x00, 0x05,
+                //  -01 LINKID type
+                0x01,
+                //  -33-00-00-26 LINKID
+                0x33, 0x00, 0x00, 0x26
+            };
+            memoryStream.Write(prefix);
+
+            var request = $"GET / HTTP/1.1\r\nHost: {serverAddress.Host}\r\nConnection: close\r\n\r\n";
+            memoryStream.Write(Encoding.ASCII.GetBytes(request));
+
+            return memoryStream.ToArray();
+        }
+    }
+}
--- a/src/ProxyProtocol/Client.Sample/Program.cs
+++ b/src/ProxyProtocol/Client.Sample/Program.cs
@@ -1,8 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.IO;
 using System.Net.Security;
 using System.Net.Sockets;
-using System.Reflection.Metadata;
 using System.Text;
 using System.Threading.Tasks;
 
@@ -12,15 +14,17 @@ namespace Client.Sample
     {
         static async Task Main(string[] args)
         {
-            var serverAddress = new Uri("https://localhost:5001");
+            var serverAddress =
+                new Uri("https://localhost:5001");
+                // new Uri("http://localhost:5000");
 
-            Console.WriteLine("Ready");
+            Console.WriteLine("Ready, press any key to run.");
             Console.ReadKey();
 
-            await ConnectAsync(serverAddress);
+            await SendRequestAsync(serverAddress);
         }
 
-        private static async Task ConnectAsync(Uri serverAddress)
+        private static async Task SendRequestAsync(Uri serverAddress)
         {
             using var client = new TcpClient();
             Console.WriteLine($"Connecting to {serverAddress}");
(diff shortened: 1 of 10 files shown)


src/ProxyProtocol/ProxyProtocol.Sample/ProxyProtocol.cs
```diff
@@ -0,0 +1,156 @@
...
+                            // It is PPv2, and we have enough data for PPv2
+                            break;
+                        }
+
+                        // It is PPv2, and we don't have enough data for PPv2
+                    }
+                }
+                finally
+                {
+                    input.AdvanceTo(buffer.Start);
```
halter73, line 79: If this receives a partial proxy preamble, it could spin indefinitely.

```suggestion
                    input.AdvanceTo(buffer.Start, buffer.End);
```

https://docs.microsoft.com/en-us/dotnet/standard/io/pipelines#pipereader-common-problems

src/ProxyProtocol/ProxyProtocol.Sample/ProxyProtocol.cs
```diff
@@ -0,0 +1,156 @@
...
+            while (true)
+            {
+                var result = await input.ReadAsync();
+                var buffer = result.Buffer;
+
+                try
+                {
+                    if (result.IsCompleted)
+                    {
+                        return;
```
halter73, line 49: Does Proxy Protocol v2 support half-open connections? I assume not. But if it does, we should inspect the buffer before exiting due to a FIN.

src/ProxyProtocol/ProxyProtocol.Sample/ProxyProtocol.cs
halter73: This also means we'll probably have to have a loop inside the try/finally that calls ExtractPpv2Data to ensure we consume everything we can before saying we've examined the entire buffere.

halter73: I now see why we didn't want to advance the examined cursor:

1. We don't want to examine past the "Ppv2Data" for proxy protocol connections.
2. We don't want to examine anything in case IsProxyProtocol later returns false in a later iteration. 

If we cannot either extract all the Ppv2Data or determine that it's not a PPv2 connection after the first read, I think we're going to have to create a new IDuplexPipe that replaces the original Reader with a new one that doesn't have everything pre-examined and set it on ConnectionContext.Transport.

src/ProxyProtocol/ProxyProtocol.Sample/ProxyProtocol.cs
```diff
@@ -0,0 +1,156 @@
...
+                    input.AdvanceTo(buffer.Start);
+                }
+            }
+
+            await next();
+        }
+
+        private static bool IsProxyProtocol(byte[] buffer)
+        {
+            if (buffer.Length > 12)
```
davidfowl, line 88: Preamble.AsSpan().SequenceEqual(buffer)

src/ProxyProtocol/ProxyProtocol.Sample/ProxyProtocol.cs
```diff
@@ -0,0 +1,156 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Connections;
+using Microsoft.Extensions.Logging;
+
+namespace ProxyProtocol.Sample
+{
+    public static class ProxyProtocol
+    {
+        // The proxy protocol marker.
+        private static readonly byte[] Preamble = { 0x0D, 0x0A, 0x0D, 0x0A, 0x00, 0x0D, 0x0A, 0x51, 0x55, 0x49, 0x54, 0x0A };
```
davidfowl, line 15: This should be types as `ReadOnlySpan<byte>`

private static readonly ReadOnlySpan<byte> Preamble => new byte[] { 0x0D, 0x0A, 0x0D, 0x0A, 0x00, 0x0D, 0x0A, 0x51, 0x55, 0x49, 0x54, 0x0A };

Tratcher: `private static ReadOnlySpan<byte> Preamble => new byte[] { 0x0D, 0x0A, 0x0D, 0x0A, 0x00, 0x0D, 0x0A, 0x51, 0x55, 0x49, 0x54, 0x0A };`

src/ProxyProtocol/ProxyProtocol.Sample/ProxyProtocol.cs
Tratcher: PPv2 is only a connection preface, the connection state isn't relevant to the protocol.

I don't think half-open is something we care about when the target audience is Kestrel HTTP scenarios.

src/ProxyProtocol/ProxyProtocol.Sample/ProxyProtocol.cs
Tratcher: I think we can advance examined so long as it's a partial match, and if it's not a partial match we can break out.

Tratcher:
@halter73 do you have time now to review this, or should we give it one more week?

src/ProxyProtocol/Client.Sample/Program.cs
```diff
@@ -0,0 +1,101 @@
...
+            Stream stream = client.GetStream();
+
+            if (serverAddress.Scheme == Uri.UriSchemeHttps)
+            {
+                Console.WriteLine("Negotiating TLS");
+                var sslStream = new SslStream(stream);
+                await sslStream.AuthenticateAsClientAsync(new SslClientAuthenticationOptions()
+                {
+                    TargetHost = serverAddress.Host,
+                    RemoteCertificateValidationCallback = (_, __, ___, ____) => true,
```
halter73, line 38: Should we be disabling normal cert validation in a sample? Trusing dev certs should be good enough, right?

src/ProxyProtocol/Client.Sample/Program.cs
```diff
@@ -0,0 +1,101 @@
...
+                stream = sslStream;
+            }
+
+            Console.WriteLine("Sending message");
+            var data = CreateMessage(serverAddress);
+            for (var i = 0; i < data.Length; i++)
+            {
+                Console.Write(".");
+                await stream.WriteAsync(data, i, 1);
+                await Task.Delay(TimeSpan.FromSeconds(0.1));
```
halter73, line 50: What's the purpose of sending the data one byte at a time?

src/ProxyProtocol/Client.Sample/Program.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            var serverAddress = new Uri("https://localhost:5001");
+
+            Console.WriteLine("Ready");
+            Console.ReadKey();
+
+            await ConnectAsync(serverAddress);
+        }
+
+        private static async Task ConnectAsync(Uri serverAddress)
```
halter73, line 23: This is doing more than connecting. It looks like this is the entire client sample. Maybe just move this into `Main` or rename it to something like `ConnectAndMakeSampleRequestAsync`.

src/ProxyProtocol/ProxyProtocol.Sample/ProxyProtocol.cs
```diff
@@ -0,0 +1,152 @@
...
+                        // Break if it is not PPv2.
+                        break;
+                    }
+
+                    if (HasEnoughPpv2Data(bufferArray))
+                    {
+                        // It is PPv2
+                        ExtractPpv2Data(ref buffer, bufferArray, connectionContext, logger);
+                        // We've consumed and sliced off the prefix.
+                        examined = buffer.Start;
```
halter73, line 84: This could move examined backwards relative to a previous call to AdvanceTo. This will cause the default PipeReader to throw as it tries to get the length of newly-examined bytes.

https://github.com/dotnet/runtime/blob/00d0dba14ed9927b22de224de94d812449529f17/src/libraries/System.IO.Pipelines/src/System/IO/Pipelines/Pipe.cs#L461

halter73: Can we just move examined to the `16 + length` evaluated in HasEnoughPpv2Data? I assume the TLS handshake cannot complete until there's more data than that.

src/ProxyProtocol/ProxyProtocol.Sample/ProxyProtocol.cs
```diff
@@ -0,0 +1,152 @@
...
+                var srcAddress = new IPAddress(srcIpAddressArray);
+
+                var destIpAddressArray = new byte[4];
+                Array.Copy(bufferArray, 20, destIpAddressArray, 0, 4);
+                var destAddress = new IPAddress(destIpAddressArray);
+
+                var srcPort = (int)(bufferArray[25] | (bufferArray[24] << 8));
+                var destPort = (int)(bufferArray[27] | (bufferArray[26] << 8));
+
+                context.Items.Add(SourceIPAddressKey, srcAddress);
```
davidfowl, line 117: Don't use Items here, use a single feature with 4 fields.

src/ProxyProtocol/Client.Sample/Program.cs
davidfowl: This looks like a test 😄

Tratcher: To test boundary conditions 😁. And yes, this is a manual test.

src/ProxyProtocol/ProxyProtocol.Sample/ProxyProtocol.cs
Tratcher: It looks like a TLS hello is way more than our 12 byte prefix minimum so rounding up shouldn't be an issue. Note this could also be used without TLS, but I wouldn't expect an issue there either, even `GET / HTTP/1.1` is more than 12 bytes.

Tratcher:
Updated and rebased.

halter73 (approved):

PR #303, opened 2020-09-02
Add experimental Blazor JS interop API

MackinnonBuck:
## Summary

This PR includes a library enabling Blazor developers to perform .NET/JS interop in a more natural way. By leveraging .NET's `DynamicObject`, JavaScript can be replaced with C# that looks and behaves almost exactly like JavaScript. For example:
```csharp
private async Task OnClick()
{
    // Get scoped access to the browser's 'window' JS object
    await using var window = JSRuntime.GetDynamicWindow();

    // Use JS-like property access and method invocation
    window.console.log("Hello, world!");

    // Create bindings to other JS objects
    var element = document.getElementById("some-element-id");

    // Mutate JS objects with ease
    element.innerText = "Not a single line of JS :)";

    // Even create your own JS objects!
    var myObject = JSObject.Create(window, new
    {
        numValue = 42,
        strValue = "37",
        objValue = new
        {
            boolValue = true
        }
    });

    // Looks like JS, acts like JS, but isn't JS (this operation is illegal in normal C#!)
    var result = myObject.numValue - myObject.strValue; // 42 - "37" = 5

    // ...
}
```

I've also added a sample Blazor WebAssembly app showing some usage examples (although a subset of the package functionality also works on Blazor Server).

## Implementation details
The `README.md` describes aspects of this package's implementation that are relevant to its usage, but here are some additional details:
* This package does not utilize the JavaScript `eval()` function. Instead, a minimal "interpreter" evaluates the expressions that the developer writes. This is done for two reasons:
  1. It keeps the implementation simple.
  2. It avoids the security risks associated with `eval()`.
* All interop calls are marshalled. When `IJSUnmarshalledRuntime` becomes part of Blazor's public API, it could be worth looking into using `InvokeUnmarshalled` for performance gains on Blazor WebAssembly.

## Future improvements
I have some ideas for how this package could be improved in the future:
* Support for JS bracket notation for property access (I'm pretty sure this could be done fairly easily).
* Support for lazily-evaluated conditional branches (`if`, `while`, etc.). This could require a lot of work.
  * Currently, using certain control flow requires the JS expression tree to be evaluated as a .NET value, which triggers an interop call.
  * This can be avoided by storing control flow in the expression tree and building both the "then" and "else" blocks on the .NET side. This obviously comes with the side effect that .NET code does not obey this type of control flow, and attempting to trigger an evaluation while building the "then" or "else" block could result in strange behavior. Furthermore, there's not a way to design this in a way that looks like JS (at least not that I know of). For these reasons, the benefits to this feature might not outweigh its quirks.
  * If implemented, it might look something like `JS.If(condition, () => { /*Build "then" block*/ }, () => { /* Build "else" block */});`
* Support for creating JS functions in .NET. Lazily-evaluated branches should be completed before implementing this, since similar challenges arise.
* As briefly discussed above, support for unmarshalled interop calls on Blazor WebAssembly.
* Allow the "root" `JSObject` to be something other than `window`. This should be easy to implement with the introduction of Blazor's `IJSObjectReference`.


wolfgangschneider:
Is this still alive? Anny plans move it to a NuGet package ?
B.R

SteveSandersonMS:
@wolfgangschneider It's an experiment that we don't have any forthcoming plans around. We'd certainly be interested in whether this turns out to be useful to you in your applications, and what makes it more useful than other JS interop patterns you might be using. If enough developers say they are using this and are being successful with it, we could consider developing it further.

PR #306, opened 2020-09-08
Ship Components.Web.Extensions preview9 package

pranavkm:
* Ship a preview9 package. Note that this targets M.A.Components 3.1 because that feels slightly better than depending on a preview version forever.
* House keeping: Remove the protected browser storage implementation


pranavkm:
This is a lot cheaper than trying to do change packaging in the ASP.NET Core repo. Plus it gets us a package that should work with preview9. The sources continue to exist in the ASP.NET Core repo, we will use that to continue shipping in 6.0.

eng/Build.props
```diff
@@ -5,8 +5,8 @@
     <ProjectToBuild Include="$(MSBuildThisFileDirectory)..\src\ComponentsElectron\ComponentsElectron.sln" />
     <ProjectToBuild Include="$(MSBuildThisFileDirectory)..\src\DotNetDiagnostics\dotnet-monitor.sln" />
     <ProjectToBuild Include="$(MSBuildThisFileDirectory)..\src\GrpcHttpApi\GrpcHttpApi.sln" />
-    <ProjectToBuild Include="$(MSBuildThisFileDirectory)..\src\ProtectedBrowserStorage\ProtectedBrowserStorage.sln" />
     <ProjectToBuild Include="$(MSBuildThisFileDirectory)..\src\Proxy\Proxy.sln" />
     <ProjectToBuild Include="$(MSBuildThisFileDirectory)..\src\WebHooks\WebHooks.sln" />
+    <ProjectToBuild Include="$(MSBuildThisFileDirectory)..\src\Components.Web.Extensions\Microsoft.AspNetCore.Components.Web.Extensions.csproj" />
```
SteveSandersonMS, line 10: I'm curious about why this needs a new top-level `src` subdirectory instead of being inside `src\Components`. Apologies if you already explained that somewhere else.

src/Components.Web.Extensions/HeadManagement/JsonSerializerOptionsProvider.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
SteveSandersonMS, line 1: Could we use the shared copy of this file from `src\Components\Shared\src\JsonSerializerOptionsProvider.cs`? Doesn't really matter if not, just wondering why not.

SteveSandersonMS (approved):
Cool!

src/Components.Web.Extensions/HeadManagement/JsonSerializerOptionsProvider.cs
pranavkm: This is in the asplabs repo. I'm using this as a way to produce a one-off version of the package for 5.0.

eng/Build.props
SteveSandersonMS: Never mind, I see this is not the same repo now!

src/Components.Web.Extensions/HeadManagement/JsonSerializerOptionsProvider.cs
SteveSandersonMS: Ah yes. That also resolves my other question.

PR #322, opened 2020-11-26
Initial commit for analyzer

pranavkm:
* Adds a sample analyzer / codefix
* Adds a WebAPI app and a "migrated" app that uses shim to allow it to be compiled. These can be used as playgrounds to test the analyzer / codefix


davidfowl:
This comment could use some more details. I see you added an ApiController base class. How much "runtime support" did you end up adding?

pranavkm:
My intent was to allow bringing over your controller types from WebAPI -> MVC with the hope that you get to compile your app after a few mechanical changes. In the absence of a shim, you get lots of error messages, and it seems like an unreasonable step to ask of people to update to remove all WebAPI references at once.

Enter the minimal compat shim. What if we can give you typings, and some of the behavior of WebAPI without tainting global state? The shim gives you the base type, and a bunch of results, enough to get your migrated controller to compile and run as an MVC app. It does not bring in any of the behavior associated with WebAPI. Those get flagged as warnings with code fixes to correct this.

More about the analyzers:

* We have an analyzer that tells you MVC does not infer HTTP method constraints from action names and provides a codefix to change that
* There's an analyzer that tells you to use attribute routes if you rely on overload resolution (and provides routes that are hopefully ambiguous). You will see a runtime error for this at Startup, so this fails fast if the analyzer messed up.
* There's an analyzer that tells you to return ResponseMessageResults instead of HttpResponseMessage. We can cheese and avoid registering a formatter.

TODO: 

- Provide fixes for WebAPI filters
- Provide help to migrate more parameter binding things like FromUriAttribute 
- Figure out how to help migrate from `HttpResponseMessage` -> `ActionResult`, `ApiController` -> `ControllerBase` etc

pranavkm:
I'm going to yolo this PR (this is labs after all

PR #368, opened 2021-07-13
Adds experimental package for Minimal APIs template

DamianEdwards:
- Update SDK, Arcade and ./eng bits
- Add initial template project for 'dotnet new api'
- For now it's the same as empty web plus a comment in `Program.cs` and the `Properties/OpenApi.cs` file to auto-configure Swashbuckle at */openapi.json* and */docs*
- Contributes to dotnet/aspnetcore#34010

``` console
Template Name             Short Name  Language  Tags
------------------------  ----------  --------  -------
ASP.NET Core Minimal API  api         [C#]      Web/API
```

### DIFF
--- a/eng/common/SetupNugetSources.ps1
+++ b/eng/common/SetupNugetSources.ps1
@@ -158,4 +158,10 @@ if ($dotnet5Source -ne $null) {
     AddPackageSource -Sources $sources -SourceName "dotnet5-internal-transport" -SourceEndPoint "https://pkgs.dev.azure.com/dnceng/internal/_packaging/dotnet5-internal-transport/nuget/v2" -Creds $creds -Username $userName -Password $Password
 }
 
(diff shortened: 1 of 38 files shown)


pranavkm (approved):

DamianEdwards:
@pranavkm any idea on the failures?

pranavkm:
Looks like the same issue as https://github.com/aspnet/AspLabs/pull/359

DamianEdwards:
@pranavkm shall I just merge and ping @JamesNK then?

pranavkm:
It's not going to produce any packages so wouldn't be super helpful to merge it in the current state. Does making the one project not packable stop the error? Because that would be an acceptable solution.

DamianEdwards:
ComponentsElectron? I can try.

DamianEdwards:
@pranavkm what did I do wrong? How am I supposed to disable the projects from packing if not via the `<IsPackable>` property?

PR #370, opened 2021-07-15
Try fix publish step

JunTaoLuo:
FYI @DamianEdwards. I'll verify this actually fixes the issue  on an internal build before merging.


eng/common/templates/post-build/channels/generic-internal-channel.yml
```diff
@@ -178,6 +178,8 @@ stages:
             /p:AzureDevOpsStaticSymbolsFeed='${{ parameters.symbolsFeed }}'
             /p:AzureDevOpsStaticSymbolsFeedKey='$(dn-bot-dnceng-artifact-feeds-rw)'
             /p:PublishToMSDL=false
+            /p:StreamingPublishingMaxClients=16
```
JunTaoLuo, line 181: I had trouble verifying this on the internal build using a branch so I'll just merge and let the official build run. If this doesn't work, I'll need ping someone from Arcade to fix it since it's an Arcade bug.

PR #374, opened 2021-07-20
Fix Swashbuckle registration in 'api' template

DamianEdwards:
- Fix issue with RoutePrefix and ordering by using rewriter to map the desired paths to the default Swashbuckle paths
- Configure Swagger UI with the desired document path (/openapi.json) so the UI shows a link to that path

### DIFF
--- a/src/ApiTemplate/templates/ApiTemplate/Properties/OpenApi.cs
+++ b/src/ApiTemplate/templates/ApiTemplate/Properties/OpenApi.cs
@@ -1,24 +1,19 @@
 using Microsoft.AspNetCore.Rewrite;
 using Microsoft.OpenApi.Models;
 
-[assembly:HostingStartup(typeof(OpenApiConfiguration))]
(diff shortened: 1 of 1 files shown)


src/ApiTemplate/templates/ApiTemplate/Properties/OpenApi.cs
```diff
@@ -1,61 +1,71 @@
...
 
-[assembly:HostingStartup(typeof(OpenApiConfiguration))]
+[assembly: HostingStartup(typeof(OpenApiConfiguration))]
 
 public class OpenApiConfiguration : IHostingStartup, IStartupFilter
 {
     private static readonly string Version = "v1";
-    private static readonly string DocumentName = "openapi.json";
+    private static readonly string DocumentPath = "openapi.json";
     private static readonly string UIPath = "docs";
```
LadyNaggaga, line 10: Thought - I know fast api uses /docs but, I think we might want to consider a different verb.

LadyNaggaga (approved):

PR #376, opened 2021-07-29
Fix parsing enums from http request to its protobuf versions

Altair7610:
I ran into a bug when converting enums from http request to protobuf. When trying to call a method from swagger containing enums in params, web api threw an exception ```InvalidOperationException: Invalid enum value '0' for enum type EnumName.```. The problem was that a comparison was performed by a string name when a number went into request.

### DIFF
--- a/src/GrpcHttpApi/src/Shared/ServiceDescriptorHelpers.cs
+++ b/src/GrpcHttpApi/src/Shared/ServiceDescriptorHelpers.cs
@@ -116,13 +116,17 @@ namespace Grpc.Shared.HttpApi
                     {
                         if (value is string s)
                         {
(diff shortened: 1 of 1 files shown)


cwe1ss:
The proto3 language guide says the following about enums in the [JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json):

> The name of the enum value as specified in proto is used. Parsers accept both enum names and integer values.

Could this support both types by using `FindValueByNumber` if `value` is a numeric type or parsable from string and `FindValueByName` otherwise?

JamesNK:
Replaced with https://github.com/aspnet/AspLabs/pull/421 which supports integers and names.

PR #382, opened 2021-09-16
Update custom elements package build for SWA compatibility

SteveSandersonMS:
The published artifact wasn't including the `.js` file. Hopefully this change to the build setup will fix that.

### DIFF
--- a/src/BlazorCustomElements/src/Microsoft.AspNetCore.Components.CustomElements/js/WebpackBuild.targets
+++ b/src/BlazorCustomElements/src/Microsoft.AspNetCore.Components.CustomElements/js/WebpackBuild.targets
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <WebpackOutputFile>$(MSBuildProjectDirectory)\wwwroot\BlazorCustomElements.js</WebpackOutputFile>
+    <WebpackOutputFile>wwwroot\BlazorCustomElements.js</WebpackOutputFile>
   </PropertyGroup>
 
   <ItemGroup>
@@ -30,6 +30,7 @@
           Outputs="$(WebpackOutputFile)">
     <Exec Command="npm run build:$(Configuration)" WorkingDirectory="$(MSBuildThisFileDirectory)" />
     <ItemGroup>
+      <Content Remove="$(WebpackOutputFile)" />
       <Content Include="$(WebpackOutputFile)" />
     </ItemGroup>
   </Target>
--- a/src/BlazorCustomElements/src/Microsoft.AspNetCore.Components.CustomElements/js/WebpackBuild.targets
+++ b/src/BlazorCustomElements/src/Microsoft.AspNetCore.Components.CustomElements/js/WebpackBuild.targets
@@ -1,4 +1,12 @@
 <Project>
+
+  <PropertyGroup>
+    <ResolveStaticWebAssetsInputsDependsOn>
+      WebpackBuild;
+      $(ResolveStaticWebAssetsInputsDependsOn)
+    </ResolveStaticWebAssetsInputsDependsOn>
+  </PropertyGroup>
+
   <ItemGroup>
     <WebpackBuildInputs Include="$(MSBuildThisFileDirectory)**" Exclude="$(MSBuildThisFileDirectory)node_modules\**" />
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
@javiercn I could really use some hints on what, if anything, I'm doing wrong here.

Before this PR, when doing `dotnet pack` locally (or `build.cmd` in the repo root), the `Microsoft.AspNetCore.Components.CustomElements.nupkg` would contain the webpack-built `.js` static web assets. But when building in CI, the generated nupkg artifact did **not** contain the SWA files.

I didn't know what I was doing wrong, but based on your earlier feedback, I made this PR to hook the webpack build onto the `ResolveStaticWebAssetsInputs` target like you suggested, hoping it would resolve the issue in CI. However, it does not. The built artifacts continue not to contain static web assets ([example](https://dev.azure.com/dnceng/public/_build/results?buildId=1366098&view=artifacts&pathAsName=false&type=publishedArtifacts)) even though when building locally it does.

Am I using SWA wrong still somehow?

javiercn:
@SteveSandersonMS there are two ways to go about this

The file are being put in wwwroot and are not being defined as content, hence they are not being picked up (unless a previous build left them there) which is why I think you are seeing it fail on CI.

There are two ways to go about this:
* Do what we do in our packages: 
 https://github.com/dotnet/aspnetcore/blob/main/src/Components/WebAssembly/Authentication.Msal/src/Microsoft.Authentication.WebAssembly.Msal.csproj
* Run your target before AssignTargetPaths and add the files to the `Content` item group.

SteveSandersonMS:
@javiercn Thanks for the advice. I tried following this [here](https://github.com/aspnet/AspLabs/pull/382/commits/0c33b1294bbf214c5bebc8a0c3169c9b0987d117), and got a really weird error:

```
Assets.targets(438,5): error : Two assets found targeting the same path with incompatible asset kinds:  [c:\git\aspnet\AspLabs\src\BlazorCustomElements\src\Microsoft.AspNetCore.Components.CustomElements\Microsoft.AspNetCore.Components.CustomElements.csproj]
C:\Program Files\dotnet\sdk\6.0.100-rc.2.21456.8\Sdks\Microsoft.NET.Sdk.Razor\targets\Microsoft.NET.Sdk.Razor.StaticWebAssets.targets(438,5): error : 'c:\git\aspnet\AspLabs\src\BlazorCustomElements\src\Microsoft.AspNetCore.Components.CustomElements\wwwroot\BlazorCustomElements.js' with kind 'All' [c:\git\aspnet\AspLabs\src\BlazorCustomElements\src\Microsoft.AspNetCore.Components.CustomElements\Microsoft.AspNetCore.Components.CustomElements.csproj]
C:\Program Files\dotnet\sdk\6.0.100-rc.2.21456.8\Sdks\Microsoft.NET.Sdk.Razor\targets\Microsoft.NET.Sdk.Razor.StaticWebAssets.targets(438,5): error : 'c:\git\aspnet\AspLabs\src\BlazorCustomElements\src\Microsoft.AspNetCore.Components.CustomElements\wwwroot\BlazorCustomElements.js' with kind 'All' [c:\git\aspnet\AspLabs\src\BlazorCustomElements\src\Microsoft.AspNetCore.Components.CustomElements\Microsoft.AspNetCore.Components.CustomElements.csproj]
```

It seems weird because it says they have incompatible asset kinds, but it also says they both have the asset kind `All`. Is this a SWA bug? Is there some other possible difference between these two that the error message isn't telling me about? To work around this, I'm also now explicitly removing the other `Content` itemgroup entry.

In [the docs](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?view=aspnetcore-6.0&tabs=visual-studio#typescript-integration-1), it doesn't say anything about having to chain onto `AssignTargetPaths` or to update the `Content` itemgroup manually. Is this because of a change in SWA for 6.0, or does the `CompileTypeScript` task somehow do it for you? It would be great to have docs that don't depend on anything specific to `CompileTypeScript` since in general people may also use other builds tools like Webpack.

SteveSandersonMS:
Update: the `"All" is not compatible with "All"` error does make sense because the rule is "you can only have one build/publish output per target path". It's only valid to have multiple content items with the same target path if only one of them gets output on build, and only one of them gets output on publish. If they are both set to "all", then they'd both be output for both, and hence conflict.

PR #383, opened 2021-09-17
[Blazor] Adds a custom packaging sample

javiercn:
This commit adds a sample that demonstrates how the blazor publish output
can be customized to enable custom bootstrapping of Blazor applications.

This commit includes two new packages 'Microsoft.AspNetCore.Components.WebAssembly.Packaging'
and 'Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server'that
work together to enable customizing the publish output for Blazor and the
subsequent customized loading process by the browser.

Microsoft.AspNetCore.Components.WebAssembly.Packaging contains MSBuild
targets to attach to the Blazor publish process and generate a single
multipart bundle file that is consumed by a JS initializer also defined
in this package and that is automatically imported by Blazor WebAssembly
when the application starts.

The JS initializer included in the package, downloads the bundle and creates
an object url for each one of the files included in the bundle that then
provides to the resource loader when asked for them.

Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server contains a
helper method to streamline the serving of the application bundle and its
compressed versions when supported, and its not technically necessary.


src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging/wwwroot/Microsoft.AspNetCore.Components.WebAssembly.Packaging.lib.module.js
```diff
@@ -0,0 +1,36 @@
...
+export async function beforeStart(wasmOptions, extensions) {
+    // Simple way of detecting we are in web assembly
+
+    if (!extensions || !extensions.multipart) {
+        return;
+    }
+
+    try {
+        const integrity = extensions.multipart['app.bundle'];
+        const bundleResponse = await fetch('app.bundle', { integrity: integrity });
```
SteveSandersonMS, line 11: Do you need `cache: no-cache` here, or some other kind of cache busting?

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging/wwwroot/Microsoft.AspNetCore.Components.WebAssembly.Packaging.lib.module.js
```diff
@@ -0,0 +1,36 @@
...
+        for (let value of bundleFromData.values()) {
+            resources.push([value, URL.createObjectURL(value)]);
+        }
+
+    } catch (error) {
+        console.log(error);
+    }
+
+    wasmOptions.loadBootResource = function (type, name, defaultUri, integrity) {
+        for (const [resource, objectUrl] of resources) {
```
SteveSandersonMS, line 22: I know the collection doesn't contain enough entries for it really to matter, but why not store it as a Map from the resource name to object URL?

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging/wwwroot/Microsoft.AspNetCore.Components.WebAssembly.Packaging.lib.module.js
```diff
@@ -0,0 +1,36 @@
...
+            if (resource.name === name) {
+                return objectUrl;
+            }
+        }
+        const res = resources;
```
SteveSandersonMS, line 27: This line seems redundant

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging/wwwroot/Microsoft.AspNetCore.Components.WebAssembly.Packaging.lib.module.js
```diff
@@ -0,0 +1,36 @@
...
+        return defaultUri;
```
SteveSandersonMS, line 28: No big deal, but you can return `null`/`undefined` to fall back on the default loader logic

src/BlazorWebAssemblyCustomInitialization/BlazorWebAssemblyCustomInitialization.sln
```diff
@@ -0,0 +1,37 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31713.378
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Components.WebAssembly.Packaging", "Microsoft.AspNetCore.Components.WebAssembly.Packaging\Microsoft.AspNetCore.Components.WebAssembly.Packaging.csproj", "{AB5B3CB1-80C9-4982-80D0-8FA482464917}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Components.WebAssembly.Packaging.Tasks", "Microsoft.AspNetCore.Components.WebAssembly.Packaging.Tasks\Microsoft.AspNetCore.Components.WebAssembly.Packaging.Tasks.csproj", "{853B5E7C-DE46-4D18-B8C9-57DC7E7B15F2}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server", "Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server\Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server.csproj", "{051ED973-E469-4F3C-BC19-65E5D05A97A5}"
```
SteveSandersonMS, line 10: I think we should have more discussion about the naming of these packages. The name `Microsoft.AspNetCore.Components.WebAssembly.Packaging` sounds very official, as if it's the official true way to package a Blazor WebAssembly application. Whereas the reality is:

 * It represents one specific bundling strategy (put everything in one file), not the idea of packaging in general
 * It's entirely optional anyway

Do you think we could find a name that's more specific to what it does? For example `Microsoft.AspNetCore.Components.WebAssembly.SingleFileBundling`?

cc @danroth27 for any thoughts about how we want to communicate this (e.g., naming, and how supported it is) to customers.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging.Tasks/BundleBlazorAssets.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        [Required]
+        public string BundlePath { get; set; }
+
+        [Output]
+        public ITaskItem[] Extension { get; set; }
+
+        public override bool Execute()
+        {
+            var bundle = new MultipartFormDataContent("--0a7e8441d64b4bf89086b85e59523b7d");
```
SteveSandersonMS, line 22: This is cool! I hadn't thought of using `MultipartFormDataContent` for this, but it saves us the trouble of a custom archive format and unpacking logic. Great idea.

src/BlazorWebAssemblyCustomInitialization/sample/CustomPackagedApp/Client/CustomPackagedApp.Client.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly" Version="6.0.0-rc.1.21423.16" />
+    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.Packaging" Version="0.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="6.0.0-rc.1.21423.16" PrivateAssets="all" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Shared\CustomPackagedApp.Shared.csproj" />
+  </ItemGroup>
+
+</Project>
```
SteveSandersonMS, line 23: It would be ideal to avoid duplicating the entire default project template here if we can. It will start to look outdated quickly as we keep changing it. Would it be sufficient to have something super minimal instead?

src/BlazorWebAssemblyCustomInitialization/sample/.local/README.txt
```diff
@@ -0,0 +1,3 @@
+To run the sample, copy the output packages from the build into this folder, which will make them available during the restore process.
+
+The packages needed are Microsoft.AspNetCore.Components.WebAssembly.Packaging and Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server
```
SteveSandersonMS, line 3: I don't personally understand these instructions clearly. Most likely I could figure it out, but let's not create homework for the customer.

I'm not sure which build it refers to . Could we either not have the sample at all, or if we are having it, give really clear instructions that don't assume the reader understands .NET build mechanisms well. Assume this is their first time using .NET.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server/BlazorPackagingApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.Builder;
+
+public static class BlazorPackagingApplicationBuilderExtensions
+{
+    public static IApplicationBuilder UseBundlingExtension(this IApplicationBuilder builder, PathString pathPrefix)
```
SteveSandersonMS, line 14: Is the idea that this does the bundling dynamically during the request instead of at build time? If so, what's the advantage of this? Who should use it and when?

There would be some advantage in limiting our offering here to just one bundling mechanism (build time) to avoid having two things to explain and maintain. Perhaps there's some reason why it's totally necessary to do it at runtime, but it's unclear to me.

SteveSandersonMS: Especially given the duplication of the content-type negotiation logic from elsewhere.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server/BlazorPackagingApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+        {
+            throw new ArgumentNullException(nameof(builder));
+        }
+
+        var webHostEnvironment = builder.ApplicationServices.GetRequiredService<IWebHostEnvironment>();
+
+        var options = CreateStaticFilesOptions(webHostEnvironment.WebRootFileProvider);
+
+        builder.MapWhen(
+            ctx => ctx.Request.Path.StartsWithSegments(pathPrefix, out var rest) && rest.Equals("/app.bundle", StringComparison.OrdinalIgnoreCase),
```
SteveSandersonMS, line 26: Is this logic equivalent to `builder.Map("/app.bundle", ...)`? Not sure on the reason for `MapWhen`. Apologies if I'm missing something obvious.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server/BlazorPackagingApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+    public static IApplicationBuilder UseBundlingExtension(this IApplicationBuilder applicationBuilder) =>
+        UseBundlingExtension(applicationBuilder, default);
+
+    private static StaticFileOptions CreateStaticFilesOptions(IFileProvider webRootFileProvider)
+    {
+        var options = new StaticFileOptions();
+        options.FileProvider = webRootFileProvider;
+        var contentTypeProvider = new FileExtensionContentTypeProvider();
+        contentTypeProvider.Mappings.Add(".bundle", "multipart/form-data; boundary=\"--0a7e8441d64b4bf89086b85e59523b7d\"");
+        contentTypeProvider.Mappings.Add(".br", "multipart/form-data; boundary=\"--0a7e8441d64b4bf89086b85e59523b7d\"");
```
SteveSandersonMS, line 45: Do we strictly need the file to be served with a special content type? Could we let the server use `application/octet-stream` or some other default, and have the JS-side logic take care of interpreting it correctly?

That way,

 * People don't need the special middleware. For example we already serve a `.blat` file as something default like `application/octet-stream` and that's already tested to work in all our hosting scenarios.
 * People serving from other web servers don't have to figure out custom config (e.g., Blazor WebAssembly Standalone on IIS - as it stands, they'd have to edit `web.config` to duplicate all these rules about MIME types, right?)

javiercn: The browser is not going to be happy if we don't provide the mime type for the content type I believe. We could get away doing more stuff on the JS side, but I want to be transparent over the network indicating that this is a multipart file.

We are doing this using the static files middleware, we could alternatively serve this from a minimal API/endpoint.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server/BlazorPackagingApplicationBuilderExtensions.cs
javiercn: Plain Map will likely work too

src/BlazorWebAssemblyCustomInitialization/sample/.local/README.txt
javiercn: These are not instructions for customers, these are instructions for people fiddling with the repo. If there's something you think I can clarify I'm happy to take suggestions.

src/BlazorWebAssemblyCustomInitialization/sample/CustomPackagedApp/Client/CustomPackagedApp.Client.csproj
javiercn: Sure, will remove a bunch of stuff from here in an additional commit. Wanted to use something like the default template so that it can be shown in the blog post I'm authoring for this.

src/BlazorWebAssemblyCustomInitialization/BlazorWebAssemblyCustomInitialization.sln
javiercn: I'm ok if we change the name, however I don't see how that's more official than other things we ship like Microsoft.AspNetCore.Components.CustomElements. I think that what makes a package official is the version stamped on it, not the name.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging/wwwroot/Microsoft.AspNetCore.Components.WebAssembly.Packaging.lib.module.js
javiercn: That's a good suggestion.

Overall I'm not trying to make this code "production ready" it's a sample people can use to check if it fixes their problems and a starting point they can use to create their own.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server/BlazorPackagingApplicationBuilderExtensions.cs
SteveSandersonMS: What about standalone static apps on IIS? We can't rely on server-side logic for those.

> I want to be transparent over the network indicating that this is a multipart file.

I would think that's a lower priority than being able to support static file hosts. I think it's possible for JS-side code to construct its own `Response` object and put in any content-type you like if you want to reinterpret an existing response, so it might not be wildly different from the code you already have. Not 100% sure what the code looks like, but this will be a mainstream scenario.

src/BlazorWebAssemblyCustomInitialization/sample/.local/README.txt
SteveSandersonMS: Like mentioned above, I don't think it's clear what it's telling someone to do, so I'd definitely spell it out more explicitly (even if it is only for people working on this repo).

src/BlazorWebAssemblyCustomInitialization/BlazorWebAssemblyCustomInitialization.sln
SteveSandersonMS: Valid point about the concept of "official", however CustomElements is at least specific to what it does. We could make these package names more specific.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging/wwwroot/Microsoft.AspNetCore.Components.WebAssembly.Packaging.lib.module.js
SteveSandersonMS: Sure, but if we blog about this, people will do it, and expect it to work without gotchas in production.

javiercn: I was actually sending no-cache from the server, I plan to add it, however I was also plan to add a disclaimer to the blog post that folks need to tune this up for their needs.

javiercn: Addressed with `no-cache`, `must-revalidate` is not a valid value.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging/wwwroot/Microsoft.AspNetCore.Components.WebAssembly.Packaging.lib.module.js
javiercn: changed to a map

src/BlazorWebAssemblyCustomInitialization/BlazorWebAssemblyCustomInitialization.sln
javiercn: Fair, what do you think about CustomPackaging or CustomPacking

src/BlazorWebAssemblyCustomInitialization/sample/CustomPackagedApp/Client/CustomPackagedApp.Client.csproj
javiercn: Reduced the sample to the minimal expression that still allows to quickly validate things work.

src/BlazorWebAssemblyCustomInitialization/sample/.local/README.txt
javiercn: I've replaced the instructions with a run script that does everything

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server/BlazorPackagingApplicationBuilderExtensions.cs
javiercn: I've removed this package in favor of using a minimal API on the sample project to serve the file

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server/BlazorPackagingApplicationBuilderExtensions.cs
javiercn: > What about standalone static apps on IIS? We can't rely on server-side logic for those.

Not trying to write a sample that covers all the casuistic, for the people that we care about, these apps are likely to be hosted. For apps that are not, they can follow the same steps to add mime types and resolve content stuff like we suggest in the deployment guide.

I did try to construct my own response and wasn't able to get it to work. I don't consider this important as the main goal of this sample is to showcase the mechanic not to be production ready nor to be a working solution for all possible scenarios.

I want to be clear that what I call success here is being able to make changes on the publish process and load the app successfully on a given environment, having a production ready solution is a non-goal (and I'll put a disclaimer in the blog post if necessary) or something that adapts to all circumstances, nor I will be providing detailed descriptions on how to setup all possible scenarios.

src/BlazorWebAssemblyCustomInitialization/BlazorWebAssemblyCustomInitialization.sln
SteveSandersonMS: CustomPacking/CustomPackaging sounds fine to me. I was going to argue again for SingleFileBundling but I think I understand your goal with the blog post is to emphasise more that this is a sample of custom packing rather than focusing on what exactly it does internally.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging/wwwroot/Microsoft.AspNetCore.Components.WebAssembly.Packaging.lib.module.js
```diff
@@ -0,0 +1,34 @@
...
+    } catch (error) {
+        console.log(error);
+    }
+
+    wasmOptions.loadBootResource = function (type, name, defaultUri, integrity) {
+        for (const [resource, objectUrl] of resources) {
+            if (resource.name === name) {
+                return objectUrl;
+            }
+        }
```
SteveSandersonMS, line 24: It's super unusual to iterate over a Map instead of keying it by the thing you want to do the lookup by :)

Totally up to you whether you want to change anything about that. I know it's only a sample and the number of entries in the map won't be huge.

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging/Microsoft.AspNetCore.Components.WebAssembly.Packaging.csproj
```diff
@@ -0,0 +1,30 @@
...
+
+  <ItemGroup>
+    <None Update="build\**" Pack="true" PackagePath="%(Identity)" />
+    <Content Include="_._" Pack="true" PackagePath="lib\net6.0\_._" />
+  </ItemGroup>
+  
+  <Target Name="GetTasksOutputDlls" BeforeTargets="CoreCompile">
+    <MSBuild Projects="..\Microsoft.AspNetCore.Components.WebAssembly.Packaging.Tasks\Microsoft.AspNetCore.Components.WebAssembly.Packaging.Tasks.csproj" Targets="Publish;PublishItemsOutputGroup" Properties="Configuration=Release">
+      <Output TaskParameter="TargetOutputs" ItemName="_TasksProjectOutputs" />
+    </MSBuild>
```
SteveSandersonMS, line 24: Does this mean it only works when building from source, and we would not publish the `.Task` project as a NuGet package? Fine with me either way, I'm just checking I understand.

src/BlazorWebAssemblyCustomInitialization/sample/.local/README.txt
SteveSandersonMS: These instructions are still here in the current PR. Is that intentional?

src/BlazorWebAssemblyCustomInitialization/sample/CustomPackagedApp/Server/Program.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    string? contentEncoding = null;
+    var contentType = "multipart/form-data; boundary=\"--0a7e8441d64b4bf89086b85e59523b7d\"";
+    var fileName = "app.bundle";
+
+    if (context.Request.Headers.AcceptEncoding.Contains("br"))
+    {
+        contentEncoding = "br";
+        fileName += ".br";
+    }else if (context.Request.Headers.AcceptEncoding.Contains("gzip"))
```
SteveSandersonMS, line 42:
```suggestion
}
else if (context.Request.Headers.AcceptEncoding.Contains("gzip"))
```

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging.Server/BlazorPackagingApplicationBuilderExtensions.cs
SteveSandersonMS: Totally understood. Your points here help to clarify the scope and goals of this sample.

We will have to phrase the blog post clearly to indicate that it showcases something you can do with the framework, rather than a turnkey solution, but I've no doubt that can be communicated well.

Even though I have been a bit fussy about clarifying what should and shouldn't work here, that shouldn't imply I think something's wrong or insufficient. This looks like a very useful sample and will no doubt be really helpful!

SteveSandersonMS (approved):
Excellent - thanks for bringing this whole work area to completion!

src/BlazorWebAssemblyCustomInitialization/Microsoft.AspNetCore.Components.WebAssembly.Packaging/Microsoft.AspNetCore.Components.WebAssembly.Packaging.csproj
javiercn: Yep, the tasks project doesn't get published, it's compiled and put into the tasks folder of the parent package. You can't really do anything with the task without the target.

src/BlazorWebAssemblyCustomInitialization/BlazorWebAssemblyCustomInitialization.sln
javiercn: Discussed with @danroth27 and ended up with `Microsoft.AspNetCore.Components.WebAssembly.MultipartBundle`

src/BlazorWebAssemblyCustomInitialization/sample/.local/README.txt
javiercn: I've updated them to point to the `run.ps1` script.

PR #385, opened 2021-09-28
Sample on how to deal with computed js files in static web assets

javiercn:


src/BlazorCustomElements/src/Microsoft.AspNetCore.Components.CustomElements/js/WebpackBuild.targets
```diff
@@ -1,38 +1,29 @@
...
-    <WebpackBuildInputs Include="$(BaseIntermediateOutputPath)webpackconfig.txt" />
-    <UpToDateCheckInput Include="@(WebpackBuildInputs)" />
+    <NpmBuildInputs Include="$(NpmFolder)**" Exclude="$(DefaultItemExcludes)" />
   </ItemGroup>
 
   <!-- If package-lock.json has changed, perform an NPM install -->
-  <Target Name="NpmInstall"
-          Inputs="$(MSBuildThisFileDirectory)package-lock.json"
-          Outputs="$(BaseIntermediateOutputPath)package-lock.json">
+  <Target Name="NpmInstall" Inputs="$(NpmFolder)package-lock.json" Outputs="$(NpmFolder)node_modules\.package-lock.json">
```
SteveSandersonMS, line 12: We still need to copy `package-lock.json` to `BaseIntermediateOutputPath` or similar if we want the incrementalism to work here

PR #387, opened 2021-10-01
ConcurrencyLimiter implementation

BrennanConroy:


src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -132,27 +165,54 @@ private void Release(int releaseCount)
...
         }
 
-        private class ConcurrencyLease : RateLimitLease
+        private void CancellationRequested(TaskCompletionSource<RateLimitLease> tcs)
         {
-            private static readonly IEnumerable<string> Empty = new string[0];
+            lock (_lock)
+            {
+                // REVIEW: failed lease or exception?
+                tcs.TrySetResult(FailedLease);
```
BrennanConroy, line 173: Should this be a OperationCanceledException/TaskCanceledException or a failed lease with a reason phrase?

halter73: We should probably capture the CancellationToken when we UnsafeRegister and either call `tcs.TrySetCanceled(token)` or `tcs.TrySetException(new OperationCanceledException(token))`. I don't have a big preference for which one. `CancellationToken.ThrowIfCancellationRequested()` throws an OperationCanceledException though, and we should probably be doing that at the start of WaitAsyncCore, so maybe we should use an OperationCanceledExceptoin for consistency.

src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -20,14 +24,20 @@ public sealed class ConcurrencyLimiter : RateLimiter
         private static readonly ConcurrencyLease SuccessfulLease = new ConcurrencyLease(true, null, 0);
         private static readonly ConcurrencyLease FailedLease = new ConcurrencyLease(false, null, 0);
 
+        /// <summary>
+        /// Initializes the <see cref="ConcurrencyLimiter"/>.
+        /// </summary>
+        /// <param name="options"></param>
```
halter73, line 30: Can we add docs for the params too?

src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -132,27 +167,53 @@ private void Release(int releaseCount)
             }
         }
 
-        private class ConcurrencyLease : RateLimitLease
+        private void CancellationRequested(TaskCompletionSource<RateLimitLease> tcs, CancellationToken token)
         {
-            private static readonly IEnumerable<string> Empty = new string[0];
+            lock (_lock)
+            {
+                tcs.TrySetException(new OperationCanceledException(token));
```
halter73, line 174: Why does this need a lock? I guess it shouldn't matter since we're using `TaskCreationOptions.RunContinuationsAsynchronously`, but it seems unnecessary.

src/RateLimiting/src/ConcurrencyLimiterOptions.cs
```diff
@@ -5,22 +5,49 @@
...
+        /// </summary>
         public int PermitLimit { get; }
 
-        // Behaviour of WaitAsync when not enough resources can be leased
-        public QueueProcessingOrder QueueProcessingOrder { get; }
+        /// <summary>
+        /// Determines the behaviour of <see cref="RateLimiter.WaitAsync"/> when not enough resources can be leased.
+        /// </summary>
+        /// <value>
+        /// <see cref="QueueProcessingOrder.OldestFirst"/>
```
halter73, line 44: I don't think `<value>` is intended to indicate just the default value unless you say that it is: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#value

```suggestion
        /// <see cref="QueueProcessingOrder.OldestFirst"/> by default.
```

src/RateLimiting/src/MetadataName.T.cs
```diff
@@ -5,43 +5,74 @@
...
         public override bool Equals([NotNullWhen(true)] object? obj)
         {
             return obj is MetadataName<T> && Equals((MetadataName<T>)obj);
         }
 
+        /// <summary>
+        /// Determines whether the specified <see cref="MetadataName{T}"/> is equal to the current object.
+        /// </summary>
+        /// <param name="other"></param>
+        /// <returns></returns>
```
halter73, line 52: Can this not be inherited from the interface?

src/RateLimiting/src/RateLimitLease.cs
```diff
@@ -27,6 +42,7 @@ public bool TryGetMetadata<T>(MetadataName<T> metadataName, [MaybeNullWhen(false
             var successful = TryGetMetadata(metadataName.Name, out var rawMetadata);
             if (successful)
             {
+                // TODO: is null metadata allowed?
                 metadata = rawMetadata is null ? default : (T)rawMetadata;
```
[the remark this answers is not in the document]
halter73: I don't recall API review saying null metadata was valid except when TryGetMetadata returns false, but the signatures of the non-generic TryGetMetadata and GetAllMetadata make me wonder.

There shouldn't be a compiler warning since we're using `[MaybeNullWhen(false)]`.

> Specifies that when a method returns ReturnValue, the parameter may be null *even if the corresponding type disallows it*.

Emphasis mine. The signature in the API proposal is `public bool TryGetMetadata<T>(MetadataName<T> metadataName, [MaybeNullWhen(false)] out T metadata);` with no `T?`.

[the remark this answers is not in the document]
halter73: I think I might remember us discussing allowing `null` to distinguish between metadata that is normally defined but not present (returns true but gives null) vs metadata that is never present (returns false and gives null). If that's the case, the signature should probably just be `public bool TryGetMetadata<T>(MetadataName<T> metadataName, [MaybeNull] out T metadata)`. This way value types don't need to be wrapped in Nullable.

src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -45,84 +58,130 @@ protected override RateLimitLease AcquireCore(int permitCount)
...
         protected override ValueTask<RateLimitLease> WaitAsyncCore(int permitCount, CancellationToken cancellationToken = default)
         {
+            cancellationToken.ThrowIfCancellationRequested();
+
             // These amounts of resources can never be acquired
-            if (permitCount < 0 || permitCount > _options.PermitLimit)
+            if (permitCount > _options.PermitLimit)
             {
-                throw new ArgumentOutOfRangeException();
+                throw new InvalidOperationException($"{permitCount} permits exceeds the permit limit of {_options.PermitLimit}.");
```
eerhardt, line 81: (Since I can't unresolve the previous conversation, starting a new one here)

I'm not sure I buy the above analysis (I know it came from API review). Looking at the [Example of ArgumentOutOfRangeException](https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception?view=net-5.0#examples) in the docs doesn't follow this reasoning. There is no property that specifies a "min age" in that example.

Same goes for the wording on https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types#argumentexception-argumentnullexception-and-argumentoutofrangeexception. 

> if bad arguments are passed to a member

Lastly, the base `RateLimiter.WaitAsync` xml doc documents the `ArgumentOutOfRangeException`. But it doesn't document `InvalidOperationException`. Following the same reasoning as above, a caller to `RateLimiter.WaitAsync` wouldn't know to catch `InvalidOperationException` if it passed in too big of a `permitCount`.

src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -45,84 +58,130 @@ protected override RateLimitLease AcquireCore(int permitCount)
...
-                    return new ValueTask<RateLimitLease>(new ConcurrencyLease(true, this, permitCount));
+                    return new ValueTask<RateLimitLease>(lease);
                 }
 
                 // Don't queue if queue limit reached
                 if (_queueCount + permitCount > _options.QueueLimit)
                 {
                     // Perf: static failed/successful value tasks?
-                    return new ValueTask<RateLimitLease>(FailedLease);
+                    return new ValueTask<RateLimitLease>(new ConcurrencyLease(false, null, 0, "Queue limit reached"));
```
eerhardt, line 102: Would it make sense to cache this `new ConcurrencyLease` in a `private static readonly` field?

src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -172,16 +249,20 @@ protected override void Dispose(bool disposing)
 
         private struct RequestRegistration
```
eerhardt, line 250: Should make this a `readonly struct`

src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -172,16 +249,20 @@ protected override void Dispose(bool disposing)
...
                 Count = requestedCount;
                 // Perf: Use AsyncOperation<TResult> instead
-                TCS = new TaskCompletionSource<RateLimitLease>();
+                TCS = tcs;
+                CancellationTokenRegistration = cancellationTokenRegistration;
             }
 
             public int Count { get; }
 
             public TaskCompletionSource<RateLimitLease> TCS { get; }
```
eerhardt, line 263: I know this was here previously, but `TCS` doesn't follow standard naming conventions.

src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -134,25 +193,43 @@ private void Release(int releaseCount)
...
+            public override IEnumerable<string> MetadataNames => Enumerable();
+
+            private IEnumerable<string> Enumerable()
+            {
+                if (_reason is null)
+                {
+                    yield break;
+                }
+
+                yield return MetadataName.ReasonPhrase.Name;
```
eerhardt, line 223:
```suggestion
                if (_reason is not null)
                {
                    yield return MetadataName.ReasonPhrase.Name;
                }
```

This will allow for future additions without having to refactor this code.

src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -45,84 +58,130 @@ protected override RateLimitLease AcquireCore(int permitCount)
...
+            }
+
+            lease = null;
+            return false;
+        }
+
         private void Release(int releaseCount)
         {
-            lock (_lock) // Check lock check
+            lock (Lock) // Check lock check
```
eerhardt, line 153: What does the comment `Check lock check` mean?

src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -45,84 +58,130 @@ protected override RateLimitLease AcquireCore(int permitCount)
...
 
                 while (_queue.Count > 0)
                 {
-                    var nextPendingRequest =
+                    RequestRegistration nextPendingRequest =
                         _options.QueueProcessingOrder == QueueProcessingOrder.OldestFirst
                         ? _queue.PeekHead()
                         : _queue.PeekTail(); 
 
                     if (GetAvailablePermits() >= nextPendingRequest.Count)
```
eerhardt, line 165: (nit) instead of calling a virtual method `GetAvailablePermits()`, you could just use `_permitCount` count here.

(I guess this applies to everywhere in this class.)

src/RateLimiting/src/ConcurrencyLimiter.cs
```diff
@@ -45,84 +58,130 @@ protected override RateLimitLease AcquireCore(int permitCount)
...
-                // Perf: static failed/successful value tasks?
                 return new ValueTask<RateLimitLease>(SuccessfulLease);
             }
 
             // Perf: Check SemaphoreSlim implementation instead of locking
-            lock (_lock) // Check lock check
+            lock (Lock) // Check lock check
             {
-                if (GetAvailablePermits() >= permitCount)
+                if (TryLeaseUnsynchronized(permitCount, out var lease))
```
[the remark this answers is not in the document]
BrennanConroy: Whoops, if you look at the rest of the file I replaced `var` usage! Just missed this :D

PR #390, opened 2021-10-07
TokenBucketRateLimiter implementation

BrennanConroy:
Depends on https://github.com/aspnet/AspLabs/pull/387 so there aren't two PRs adding the test infra/csproj properties.

### DIFF
--- a/src/RateLimiting/src/TokenBucketRateLimiter.cs
+++ b/src/RateLimiting/src/TokenBucketRateLimiter.cs
@@ -17,7 +17,7 @@ namespace System.Threading.RateLimiting
     {
         private int _tokenCount;
         private int _queueCount;
-        private long _lastReplenishmentTick = Environment.TickCount;
+        private int _lastReplenishmentTick = Environment.TickCount;
 
         private readonly Timer? _renewTimer;
         private readonly TokenBucketRateLimiterOptions _options;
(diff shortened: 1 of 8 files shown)


src/RateLimiting/test/ConcurrencyLimiterTests.cs
```diff
@@ -105,6 +105,28 @@ public async Task FailsWhenQueuingMoreThanLimit()
             Assert.False(failedLease.IsAcquired);
         }
 
+        [Fact]
+        public async Task QueueAvailableAfterQueueLimitHitAndResources_BecomeAvailable()
```
BrennanConroy, line 109: Noticed missing test coverage when implementing the token bucket rate limiter so added this test for both.

src/RateLimiting/src/TokenBucketRateLimiter.cs
```diff
@@ -123,66 +157,88 @@ public bool TryReplenish()
...
-            {
-                return;
-            }
+            TokenBucketRateLimiter limiter = (state as TokenBucketRateLimiter)!;
+            Debug.Assert(limiter is not null);
 
-            var limiter = (TokenBucketRateLimiter)state;
+            // TODO: Handle wrapping? TickCount will wrap after ~24.8 days
+            // Use Environment.TickCount instead of DateTime.UtcNow to avoid issues on systems where the clock can change
+            long nowTicks = Environment.TickCount * TimeSpan.TicksPerMillisecond;
```
halter73, line 165: `Environment.TickCount64`? We're only getting millisecond precision to begin with, so why are we converting to TimeSpan ticks?

src/RateLimiting/src/TokenBucketRateLimiter.cs
```diff
@@ -4,16 +4,19 @@
...
-#pragma warning disable 1591
+    /// <summary>
+    /// <see cref="RateLimiter"/> implementation that replenishes tokens periodically instead of via a release mechanism.
+    /// </summary>
     public sealed class TokenBucketRateLimiter : RateLimiter
     {
         private int _tokenCount;
         private int _queueCount;
-        private long _lastReplenishmentTick;
+        private long _lastReplenishmentTick = Environment.TickCount;
```
halter73, line 19:
```suggestion
        private long _lastReplenishmentTick = Environment.TickCount * TimeSpan.TicksPerMillisecond;
```

Though I think we should just use `Environment.TickCount64` and not multiply by 10,000 everywhere.

halter73 (commented):
Let's file an issue for making rate limiters disposable.

halter73 (approved):

PR #405, opened 2021-12-12
Allow spaces in outputpath

MariovanZeist:
fixes: https://github.com/dotnet/aspnetcore/issues/38702

Assume the following:

when using the following scripts in ```package.json```
```
  "scripts": {
    "build": "npm run build:release",
    "build:Release": "webpack --mode production",
    "build:Debug": "webpack --mode development",
  }
```

When there are spaces in the output path, 

The build output shows:
`webpack --mode development "--output-path"
"D:\My" "Projects\Redacted\src\MyLibrary.Components\obj\Debug\net6.0\clientassets/"`

but should say:
`webpack --mode development --output-path
"D:\My Projects\Redacted\src\MyLibrary.Components\obj\Debug\net6.0\clientassets/"`

(for readability I put the path part on the next line)

This  PR fixes that.

### DIFF
--- a/src/ClientAssets/Microsoft.AspNetCore.ClientAssets/build/netstandard2.0/Microsoft.AspNetCore.ClientAssets.targets
+++ b/src/ClientAssets/Microsoft.AspNetCore.ClientAssets/build/netstandard2.0/Microsoft.AspNetCore.ClientAssets.targets
@@ -32,7 +32,7 @@
     </PropertyGroup>
 
     <MakeDir Directories="$(_ClientAssetsOutputFullPath)" />
-    <Exec Command="$(ClientAssetsBuildCommand) -- $(ClientAssetsBuildOutputParameter) $(_ClientAssetsOutputFullPath)" WorkingDirectory="$(ClientAssetsDirectory)" />
+    <Exec Command="$(ClientAssetsBuildCommand) -- $(ClientAssetsBuildOutputParameter) '$(_ClientAssetsOutputFullPath)'" WorkingDirectory="$(ClientAssetsDirectory)" />
 
     <ItemGroup>
       <_ClientAssetsBuildOutput Include="$(IntermediateOutputPath)clientassets\**"></_ClientAssetsBuildOutput>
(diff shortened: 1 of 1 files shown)


src/ClientAssets/Microsoft.AspNetCore.ClientAssets/build/netstandard2.0/Microsoft.AspNetCore.ClientAssets.targets
```diff
@@ -32,7 +32,7 @@
     </PropertyGroup>
 
     <MakeDir Directories="$(_ClientAssetsOutputFullPath)" />
-    <Exec Command="$(ClientAssetsBuildCommand) -- $(ClientAssetsBuildOutputParameter) $(_ClientAssetsOutputFullPath)" WorkingDirectory="$(ClientAssetsDirectory)" />
+    <Exec Command="$(ClientAssetsBuildCommand) -- $(ClientAssetsBuildOutputParameter) &quot;$(_ClientAssetsOutputFullPath)&quot;" WorkingDirectory="$(ClientAssetsDirectory)" />
```
TanayParikh, line 35: Does the following work?

```suggestion
    <Exec Command="$(ClientAssetsBuildCommand) -- $(ClientAssetsBuildOutputParameter) '$(_ClientAssetsOutputFullPath)'" WorkingDirectory="$(ClientAssetsDirectory)" />
```

MariovanZeist: For some reason, it wants the ``quot;`` instead of a real quote.  because a single quote will again result in an error.

I am not sure if this is a webpack "thing", but I must admit the fact that my "solution" works is odd.

I prepared a minimal repro https://github.com/MariovanZeist/AssetsInSpace to show the problem.

Try building this project from a folder without spaces in them, and again with spaces in them and see the difference.

MariovanZeist: @TanayParikh I did a little more research and found that besides using ``quot; `` escaping with ``%22`` also works.
The closest to official MS documentation on this I got was https://docs.microsoft.com/en-us/visualstudio/msbuild/special-characters-to-escape?view=vs-2022 and especially the note on that page stating:
> In some scenarios, you may need to escape double quote (") characters, such as when using within an Exec task

It's not actually ``MSBuild`` that is failing, but the executable ``npm`` that it called with the ``Exec`` task. It interpreted spaces as separate commandos.

MariovanZeist: @TanayParikh I was under the mistaken assumption this package would be used for production, but @javiercn cleared it up for me https://github.com/dotnet/aspnetcore/issues/38445#issuecomment-1196549022

So for me, this problem is resolved.

PR #415, opened 2022-02-01
Outline for System.Web adapters

pranavkm:


activate.ps1
```diff
@@ -0,0 +1,66 @@
+#
+# This file must be used by invoking ". .\activate.ps1" from the command line.
```
twsouthwick, line 2: What are these activate/deactivate.ps1 files for?

src/SystemWebAdapters/src/System.Web.Adapters.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
twsouthwick, line 4: What TFMs are we going to support? We had discussed at one point making it support .NET Standard which would be really helpful for transitioning libraries, but this is definitely fine to start with.

twsouthwick (approved):

activate.ps1
pranavkm: They're build infrastructure from dotnet/aspnetcore that were missing here. It allows using the dotnet SDK that gets restored as part of build.

src/SystemWebAdapters/src/System.Web.Adapters.csproj
pranavkm: I'm not sure that would work - users aren't going to be able to use this package in a netstandard2.0 library and expect it to work in a .NET Fx app because this does not substitute for System.Web's implementation.

twsouthwick: Good point. If there's a way to enable that, it would be awesome, but not sure how to ensure those types get unified.

twsouthwick:
Can you add a sln file for this?

src/SystemWebAdapters/src/SameSiteMode.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace System.Web
+{
+    public class SameSiteMode
```
twsouthwick, line 3: This should be an enum

PR #420, opened 2022-02-04
Update adapters README with high level example

twsouthwick:


twsouthwick:
Building on #415

twsouthwick:
@davidfowl Here's a high level overview

src/SystemWebAdapters/README.md
```diff
@@ -1,3 +1,61 @@
...
+
+## Supported Targets
+- .NET Core App 3.1: This will implement the adapters against ASP.NET Core `HttpContext`. This will provide the following:
+  - Conversions between ASP.NET Core `HttpContext` and `System.Web` adapter `HttpContext` (with appropriate caching so it will not cause perf hits for GC allocations)
+  - Default implementations against `Microsoft.AspNetCore.Http.HttpContext`
+  - Services that can be implemented to override some functionality such as session/caching/etc that may need to be customized to match experience.
+- .NET Standard 2.0: This will essentially be a reference assembly. There will be no constructors for the types as ASP.NET Core will construct them based on their `HttpContext` and on framework there are already other constructors. However, this will allow class libraries to target .NET Standard instead of needing to multi-target which will then require everything it depends on to multi-target.
+- .NET Framework 4.7.2: This will type forward the adapter classes to `System.Web` so that they can be unified and enable libraries built against .NET Standard 2.0 to run on .NET Framework instances.
```
davidfowl, line 61: This is what I think we need to understand better. We're producing a System.Web package that tries to type forward to the real one on .NET Framework?

twsouthwick: Yup that's the goal. It will greatly simplify how customers can build libraries. The challenge I can see here is that the .NET Standard one will mostly be no-ops (essentially a reference assembly).

PR #425, opened 2022-02-09
Add additional APIs for HttpContext adapters

twsouthwick:
Adds additional API surface area for HttpContext adapters that are highly prevalent in user code and NuGet packages.


pranavkm (approved):

src/SystemWebAdapters/src/HttpSessionStateBase.cs
```diff
@@ -0,0 +1,59 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections;
+using System.Collections.Specialized;
+
+namespace System.Web
+{
+    public abstract class HttpSessionStateBase : ICollection, IEnumerable
```
pranavkm, line 9: I'm a little wary about having this at least until we know there's a legitimate way for us to support session.

twsouthwick: Sounds good. I've removed it for now. We can add it back later

PR #427, opened 2022-02-10
Implement HttpContext against Core HttpContext

twsouthwick:
This change starts plumbing ASP.NET Core HttpContext through the HttpContext
and sets up the caching mechanism used to ensure only a single instance
of the HttpContext is created per context.

Conversion between the adapters and core types are provided via both
implicit operators as well as extension methods that get or unwrap the
adapters. The HttpContext adapter is stored in the core
HttpContext.Features so that only a single instance is used for each
instance of the core context. The additional members (such as .Request
or .Response) are cached within the adapters and only created on demand.

This change also includes the HttpContextWrapper, HttpRequestWrapper,
and HttpResponseWrapper which are converted and stored in a similar way
as to the HttpContext adapter.


twsouthwick:
/cc @davidfowl  @pranavkm 

I'm especially interested in your thoughts of storing the adapter in `HttpContext.Features`. I could swap it to use `.Items`, but this seems like it would have a bit better perf for those who are using it often since `.Items` is built on the `.Features`.

src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+            if (result is null)
+            {
+                result = new HttpContextWrapper(context);
+                context.Features.Set(result);
+            }
+
+            return result!;
+        }
+
+        [return: NotNullIfNotNull("request")]
```
davidfowl, line 51: Why do we need these?

twsouthwick: That was to provide support for `adapter.UnwrapAdapter()` and `context.GetAdapter()` to have nullability information flow correctly if the objects calling it were null.

davidfowl:
> I'm especially interested in your thoughts of storing the adapter in HttpContext.Features. I could swap it to use .Items, but this seems like it would have a bit better perf for those who are using it often since .Items is built on the .Features.

This is good (this was my idea as well).
FeelMeThinkAboutItGIF (2)

src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
davidfowl: But why do we need these public

twsouthwick: Pushed a change to move them internal

PR #434, opened 2022-02-14
Implement HttpContext.Items against core

twsouthwick:


src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
```diff
@@ -75,5 +78,11 @@ public static void AddSystemWebAdapters(this IServiceCollection services)
 
         [return: NotNullIfNotNull("response")]
         internal static HttpResponseCore? UnwrapAdapter(this HttpResponse? response) => response;
+
+        internal static IDictionary AsNonGeneric(this IDictionary<object, object?> dictionary)
+         => dictionary is IDictionary d ? d : new NonGenericDictionaryWrapper(dictionary);
```
pranavkm, line 83:
```suggestion
             => dictionary is IDictionary d ? d : new NonGenericDictionaryWrapper(dictionary);
```

src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
```diff
@@ -75,5 +78,11 @@ public static void AddSystemWebAdapters(this IServiceCollection services)
...
+
+        internal static ICollection AsNonGeneric<T>(this ICollection<T> collection)
```
pranavkm, line 85: Where do you use this?

src/SystemWebAdapters/src/HttpContext.cs
```diff
@@ -58,7 +59,18 @@ public HttpResponse Response
...
         }
 
-        public IDictionary Items => throw new NotImplementedException();
+        public IDictionary Items
+        {
+            get
+            {
+                if (_items is null)
+                {
+                    _items = _context.Items.AsNonGeneric();
```
pranavkm, line 68: I was trying to rationalize why we would create a wrapper over `HttpContext.Items` instead of returning a separate dictionary instance. I suppose this lets you stash state in the as part of a middleware and read it as part of the System.Web code which would be desirable.

src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
pranavkm: Should we consistently wrap it all the time? We can't really rationalize the behavior of any of the methods in the original implementation. For e.g if `HttpContext.Items` is implemented by a `Dictionary<object, object>` where the indexer throws if a KeyNotFoundException, it would be really confusing to users.

src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
twsouthwick: This is for the Keys/ValueCollection to check if it already implements ICollection. If we go to simply always wrapping, then I can remove it.

src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
twsouthwick: The System.Web Items is backed by a Hashtable, which does not throw in that case (I specifically handle that case currently for the wrapper. The `IDictionary.this[object]` implementation for Dictionary<,> doesn't throw, so I went with this since it seemed to be the same behavior but minimized allocations.

That said, it may not be a big deal to create the wrapper regardless to ensure consistent behavior. Let me know if you have a strong opinion there and I'm fine changing it.

src/SystemWebAdapters/src/HttpContext.cs
twsouthwick: Yes, that was feedback from one of the teams of keeping the two worlds in sync. Hence, the wrapping here.

src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
pranavkm: Sorry, I meant if `HttpContext.Items { get; set; } = new Dictionary<object, object?>();` this would be incorrect, no?

twsouthwick: It would work - see the implementation here: https://source.dot.net/#System.Private.CoreLib/Dictionary.cs,1309. The indexer is explicitly defined for `IDictionary` to handle the not found case.

twsouthwick: @pranavkm Would you like me to always wrap them or is this ok given the explicit implementation and docs?

pranavkm: Let's keep this as-is for now. Should be a small cost to revisit this later on.

PR #435, opened 2022-02-14
Implement some HttpRequest APIs against core

twsouthwick:
This implements most of the APIs except the request body and the collection types (i.e. headers and cookies).


src/SystemWebAdapters/src/SystemWebStartupFilter.cs
```diff
@@ -0,0 +1,27 @@
...
+    internal class SystemWebStartupFilter : IStartupFilter
+    {
+        public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
+            => builder =>
+            {
+                builder.Use(RegisterRawUrl);
+                next(builder);
+            };
+
+        // RawUrl from System.Web was the original path requested before anything else may have changed it. This method will cache that early on so it can be used if needed.
```
pranavkm, line 20: I personally wouldn't fret too much about something in the request line modifying the request url and just do it at the callsite.

pranavkm (approved):
Most of this looks fine. `HttpRequest.IsLocal` looks a little iffy, it would be good to get a sign off from @Tratcher before we merge it.

src/SystemWebAdapters/src/HttpRequest.cs
```diff
@@ -4,81 +4,153 @@
...
-        public string Path => throw new NotImplementedException();
+        public string? Path => _request.Path.Value;
 
         public NameValueCollection Headers => throw new NotImplementedException();
 
-        public Uri Url => throw new NotImplementedException();
+        public Uri Url => new(_request.GetEncodedUrl());
 
-        public string RawUrl => throw new NotImplementedException();
+        public string RawUrl => _request.HttpContext.GetMetadata().RawUrl ?? throw new InvalidOperationException(LocalizedStrings.RawUrlNotAvailable);
```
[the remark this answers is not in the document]
twsouthwick: What do you normally name the resource file then? Or do you just put the string error message in the code?

src/SystemWebAdapters/src/HttpRequest.cs
```diff
@@ -4,81 +4,153 @@
...
+        public string UserHostAddress => _request.Host.Value;
 
-        public string[] UserLanguages => throw new NotImplementedException();
+        public string[] UserLanguages
+        {
+            get
+            {
+                if (_userLanguages is null)
+                {
+                    _userLanguages = TypedHeaders.AcceptLanguage
```
davidfowl, line 47: Can we de linq this?

src/SystemWebAdapters/src/HttpRequest.cs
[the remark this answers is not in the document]
pranavkm: > Or do you just put the string error message in the code?

Yup: https://github.com/dotnet/aspnetcore/blob/main/src/DefaultBuilder/src/WebApplication.cs#L206

src/SystemWebAdapters/src/HttpRequest.cs
```diff
@@ -1,89 +1,202 @@
...
-        public string Path => throw new NotImplementedException();
+        public string? Path => _request.Path.Value;
 
         public NameValueCollection Headers => throw new NotImplementedException();
 
-        public Uri Url => throw new NotImplementedException();
+        public Uri Url => new(_request.GetEncodedUrl());
 
-        public string RawUrl => throw new NotImplementedException();
+        public string RawUrl => Uri.UnescapeDataString(_request.GetDisplayUrl());
```
Tratcher, line 36: Best to leave this as NotImplemented until we get a more accurate implementation.

Getting the equivalent value here is a lot more complex. It's not just 'what did the client send' because the client doesn't send it in a url format. Http.Sys makes a url for its representation of the request.

You can start with `_request.HttpContext.Features.Get<IHttpRequestFeature>().RawTarget`, but that won't have the scheme or host most of the time. It will take some parsing to make sure, and to add them. Even then, I think Http.Sys applied some basic formatting to this value so it wasn't the raw client input.

src/SystemWebAdapters/src/HttpRequest.cs
```diff
@@ -1,89 +1,202 @@
...
 
-        public string HttpMethod => throw new NotImplementedException();
+        public string HttpMethod => _request.Method;
 
-        public string UserHostAddress => throw new NotImplementedException();
+        public string UserHostAddress => _request.Host.Value;
```
Tratcher, line 40: _request.HttpContext.Connection.RemoteIpAddress?.ToString()

src/SystemWebAdapters/src/HttpRequest.cs
```diff
@@ -1,89 +1,200 @@
...
+                    if (length == 0)
+                    {
+                        _userLanguages = Array.Empty<string>();
+                    }
+                    else
+                    {
+                        var qualityArray = ArrayPool<StringWithQualityHeaderValue>.Shared.Rent(length);
+                        var userLanguages = new string[length];
+
+                        try
```
davidfowl, line 60: Nuke the try finally.

src/SystemWebAdapters/src/HttpRequest.cs
```diff
@@ -1,89 +1,200 @@
...
+        public string AppRelativeCurrentExecutionFilePath => $"~{_request.Path}";
 
-        public string ApplicationPath => throw new NotImplementedException();
+        public string ApplicationPath => _request.PathBase;
 
-        public Uri UrlReferrer => throw new NotImplementedException();
+        public Uri? UrlReferrer => TypedHeaders.Referer;
 
-        public int TotalBytes => throw new NotImplementedException();
+        public int TotalBytes => (int)_request.ContentLength.GetValueOrDefault();
```
Tratcher, line 152: Since System.Web buffered by default, TotalBytes probably also worked for Chunked requests. We'll want to revisit this when we look at the request body buffering.

davidfowl: This will require buffering that's right.

twsouthwick: Added a comment to the code

src/SystemWebAdapters/src/HttpRequest.cs
twsouthwick: done

src/SystemWebAdapters/src/HttpRequest.cs
twsouthwick: Ok, I've marked it as not implemented with a link to this thread. Once we get the low hanging fruit, we can see what is left.

src/SystemWebAdapters/src/HttpRequest.cs
twsouthwick: done

src/SystemWebAdapters/src/HttpRequest.cs
```diff
@@ -1,89 +1,197 @@
...
+        }
 
-        public string AppRelativeCurrentExecutionFilePath => throw new NotImplementedException();
+        /// <summary>
+        /// Same as CurrentExecutionFilePath, but made relative to the application root, so it is application-agnostic.
+        /// </summary>
+        /// <remarks>
+        /// See https://stackoverflow.com/questions/9701309/get-app-relative-url-from-request-url-absolutepath.
+        /// </remarks>
+        public string AppRelativeCurrentExecutionFilePath => $"~{_request.Path}";
```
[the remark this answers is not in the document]
twsouthwick: Can you give an example of what the difference here would be?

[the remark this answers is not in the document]
twsouthwick: Ah such as `/B/ C` would be `/B/%20C` with `_request.Path` but `/B/ C` with `_request.Path.Value`. Updated the test to cover that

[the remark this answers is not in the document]
twsouthwick: Updated

src/SystemWebAdapters/src/HttpRequest.cs
```diff
@@ -1,89 +1,186 @@
...
+        /// <summary>
+        /// Same as CurrentExecutionFilePath, but made relative to the application root, so it is application-agnostic.
+        /// </summary>
+        /// <remarks>
+        /// See https://stackoverflow.com/questions/9701309/get-app-relative-url-from-request-url-absolutepath.
+        /// </remarks>
+        public string AppRelativeCurrentExecutionFilePath => $"~{_request.Path.Value}";
 
-        public string ApplicationPath => throw new NotImplementedException();
+        public string ApplicationPath => _request.PathBase;
```
Tratcher, line 133: This one too:
```suggestion
        public string ApplicationPath => _request.PathBase.Value;
```

twsouthwick: Applied with test update

PR #437, opened 2022-02-16
Implement header wrapper for HttpRequest and HttpResponse

twsouthwick:


src/SystemWebAdapters/src/Internal/StringValuesNameValueCollection.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        public override void Set(string? name, string? value)
+        {
+            if (name is null)
+            {
+                return;
+            }
+
+            _headers.Remove(name);
```
Tratcher, line 74: Not needed.
```suggestion
```

Tratcher (approved):

src/SystemWebAdapters/src/Internal/StringValuesNameValueCollection.cs
twsouthwick: Good catch  :) It already overwrites it by setting the value!

PR #447, opened 2022-03-08
Buffer input stream for System.Web adapters

twsouthwick:
This sets up the input stream to be buffered and reads it all into memory to mimic the behavior of System.Web. This can be turned on per endpoint via the `SystemWebAdapterAttribute` metadata entry.

A WebApi endpoint has also been added to test similar behavior between the two frameworks that can be expanded as more features are brought online.


twsouthwick:
@davidfowl @Tratcher I haven't done much with endpoint metadata, but this looks like a good way to allow configurability per endpoint for System.Web compatibility. I'm assuming "last one wins" is a decent enough way to manage it and I don't need to deal with merging multiple layers of metadata (i.e. through the GetOrderedMetadata<> call).

As always, feedback is appreciated!

src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
```diff
@@ -14,8 +16,23 @@ public static class SystemWebAdaptersExtensions
...
             services.AddHttpContextAccessor();
+            services.AddSingleton<SystemWebAdapterMiddleware>();
         }
 
+        public static void UseSystemWebAdapters(this IApplicationBuilder app)
+        {
+            app.UseMiddleware<SystemWebAdapterMiddleware>();
+        }
+
+        public static TBuilder RequireSystemWebAdapter<TBuilder>(this TBuilder builder, SystemWebAdapterAttribute? options = null)
```
davidfowl, line 27: Do you plan to put all the functionality under this metadata?

src/SystemWebAdapters/src/Internal/SystemWebAdapterMiddleware.cs
```diff
@@ -0,0 +1,59 @@
...
+                {
+                }
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(bytes);
+            }
+
+            context.Request.Body.Position = 0;
+        }
```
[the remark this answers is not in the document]
twsouthwick: Cool! Didn't know about this method - much easier to read :)

src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
twsouthwick: I just want a way to tweak some of the behavior so that users don't need to take the full cost of i.e. buffering/draining the input stream if they find they don't need it for a specific endpoint. I'm currently seeing things as an off/on, so set it up with a single poco with properties. I'm not tied to that, and could see it a bit easier to compose if things were individual metadata items, but then there's more of a cost per request to lookup.

src/SystemWebAdapters/src/HttpRequest.cs
```diff
@@ -96,7 +96,9 @@ public string[] UserLanguages
             set => _request.ContentType = value;
         }
 
-        public Stream InputStream => throw new NotImplementedException();
+        public Stream InputStream => _request.Body.CanSeek
+            ? _request.Body
+            : throw new InvalidOperationException("Input stream must be seekable. Ensure you are calling IApplicationBuilder.UseSystemWebAdapters in your startup configuration");
```
Tratcher, line 101: I have to wonder how many apps do not rely on seeking. If you always returned _request.Body here, apps that do rely on seeking would still get a seek error, just without the extra helper text.

twsouthwick: True, but checking early will help identify what they should do

src/SystemWebAdapters/src/Middleware/BufferRequestStreamMiddleware.cs
```diff
@@ -0,0 +1,44 @@
...
+    public BufferRequestStreamMiddleware(ILogger<BufferRequestStreamMiddleware> logger)
+    {
+        _logger = logger;
+    }
+
+    public async Task InvokeAsync(HttpContextCore context, RequestDelegate next)
+    {
+        var metadata = context.GetEndpoint()?.Metadata.GetMetadata<IBufferRequestStreamMetadata>();
+
+        if (metadata is not null && metadata.IsEnabled)
```
davidfowl, line 25: C# 10 is calling you...

```suggestion
        if (metadata is { IsEnabled: true })
```

davidfowl, line 25: OR MOAR C# 10

```suggestion
        if (context.GetEndpoint()?.Metadata.GetMetadata<IBufferRequestStreamMetadata>() { IsEnabled: true } metadata)
```

src/SystemWebAdapters/src/Middleware/BufferRequestStreamMiddleware.cs
```diff
@@ -0,0 +1,44 @@
...
+            if (metadata.BufferLimit.HasValue)
+            {
+                context.Request.EnableBuffering(metadata.BufferThreshold, metadata.BufferLimit.Value);
+            }
+            else
+            {
+                context.Request.EnableBuffering(metadata.BufferThreshold);
+            }
+
+            await context.Request.Body.DrainAsync(context.RequestAborted);
```
Tratcher, line 38: @davidfowl, what do we do about IHttpMaxRequestBodySizeFeature? See https://github.com/dotnet/aspnetcore/pull/40279

src/SystemWebAdapters/src/Middleware/BufferRequestStreamMiddleware.cs
twsouthwick: That's cool - haven't used this newest pattern matching

src/SystemWebAdapters/src/BufferRequestStreamAttribute.cs
```diff
@@ -0,0 +1,19 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Web.Metadata;
+
+namespace System.Web;
```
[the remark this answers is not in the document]
twsouthwick: I'm fine with that. I was playing around with some namespacing for the metadata, but I kind of like keeping everything *new* separated so it's a bit clearer

src/SystemWebAdapters/src/SystemWebAdaptersExtensions.cs
```diff
@@ -14,8 +17,21 @@ public static class SystemWebAdaptersExtensions
         public static void AddSystemWebAdapters(this IServiceCollection services)
         {
             services.AddHttpContextAccessor();
+            services.AddSingleton<BufferRequestStreamMiddleware>();
```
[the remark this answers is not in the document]
twsouthwick: Cool - didn't realize it would by dynamically activated

src/SystemWebAdapters/src/Middleware/BufferRequestStreamMiddleware.cs
davidfowl: Pattern matching fTW

src/SystemWebAdapters/src/Adapters/PreBufferRequestStreamMiddleware.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            return PreBufferAsync(context, metadata, next);
+        }
+
+        return next(context);
+    }
+
+    private async Task PreBufferAsync(HttpContextCore context, IPreBufferRequestStreamMetadata metadata, RequestDelegate next)
+    {
+        var feature = context.Features.Get<IHttpMaxRequestBodySizeFeature>();
```
Tratcher, line 33: The buffer limit doesn't need to be related to IHttpMaxRequestBodySizeFeature.MaxRequestBodySize, the server will enforce MaxRequestBodySize independently. What David and I need to decide is if this middleware should read RequestSizeLimitAttribute and set MaxRequestBodySize.

https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/RequestSizeLimitAttribute.cs

src/SystemWebAdapters/src/Adapters/SystemWebAdaptersExtensions.cs
```diff
@@ -16,6 +17,18 @@ public static void AddSystemWebAdapters(this IServiceCollection services)
...
 
+        public static void UseSystemWebAdapters(this IApplicationBuilder app)
+        {
+            app.UseMiddleware<PreBufferRequestStreamMiddleware>();
+        }
+
+        /// <summary>
+        /// Adds request stream bufferingto the endpoint(s)
+        /// </summary>
+        public static TBuilder RequireRequestStreamBuffering<TBuilder>(this TBuilder builder, IPreBufferRequestStreamMetadata? metadata = null)
```
Tratcher, line 28:
```suggestion
        public static TBuilder PreBufferRequestStream<TBuilder>(this TBuilder builder, IPreBufferRequestStreamMetadata? metadata = null)
```

src/SystemWebAdapters/src/Adapters/PreBufferRequestStreamMiddleware.cs
twsouthwick: Ah gotcha. Is that something we can decide on later then so it doesn't block here? I can put a comment in the code for that

twsouthwick: I've reverted to just my metadata but added a comment with a link to this discussion

twsouthwick:
@Tratcher I had to put back the `services.AddSingleton<PreBufferRequestStreamMiddleware>()`. Without that, I get an error about it not being registered. Maybe there's a different pattern you're thinking of?

Tratcher (approved):

PR #454, opened 2022-03-16
Set FormatEnumsAsIntegers to true by default

jgreer013:
Enums should be formatted as integers by default, as deserializing string-based enums can cause exceptions when the value names (not the values themselves) are changed, or when new values are added (unlike integers, which are handled gracefully on deserialization generally).


jgreer013:
I realize this might have ramifications for consumers of this package, but I think it makes sense to make this change, just not sure when.

PR #461, opened 2022-03-24
Use convention based middleware instead of IMiddleware

twsouthwick:


twsouthwick:
@Tratcher you mentioned this was the style more generally used. Since these middleware can be singletons, is there any concern with creating new ones with each request?

Tratcher (approved):
> @Tratcher you mentioned this was the style more generally used. Since these middleware can be singletons, is there any concern with creating new ones with each request?

Even with this pattern the middleware isn't created per request, it's created once at startup.

PR #463, opened 2022-03-25
Add runtime helper for migration content root

twsouthwick:
The migration tooling when launched will have the content root by default show up as the framework app. The content root is passed in as a relative path which we need to handle before passing into WebApplicationBuilder

### DIFF
--- a/src/SystemWebAdapters/samples/MvcCoreApp/Program.cs
+++ b/src/SystemWebAdapters/samples/MvcCoreApp/Program.cs
@@ -1,6 +1,7 @@
 using System.Web.Adapters;
 
-var builder = WebApplication.CreateBuilder(args);
+var builder = WebApplication.CreateBuilder(MigrationWebApplicationOptions.CreateWithRelativeContentRoute(args));
+builder.Services.AddReverseProxy().LoadFromConfig(builder.Configuration.GetSection("ReverseProxy"));
 
 // Add services to the container.
 builder.Services.AddControllersWithViews();
@@ -25,8 +26,10 @@ app.UseAuthorization();
 
 app.UseSystemWebAdapters();
 
-app.MapControllerRoute(
-    name: "default",
-    pattern: "{controller=Home}/{action=Index}/{id?}");
+app.UseEndpoints(endpoints =>
+{
+    endpoints.MapDefaultControllerRoute();
+    endpoints.MapReverseProxy();
+});
 
 app.Run();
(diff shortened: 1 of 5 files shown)


src/SystemWebAdapters/src/Adapters/MigrationWebApplicationOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+
+namespace System.Web.Adapters;
+
+public class MigrationWebApplicationOptions
+{
+#if NET6_0_OR_GREATER
+    public static WebApplicationOptions CreateWithRelativeContentRoute(string[] args)
```
davidfowl, line 15: This API is a little ugly...

src/SystemWebAdapters/samples/MvcCoreApp/Program.cs
```diff
@@ -25,8 +26,10 @@
...
 app.UseSystemWebAdapters();
 
-app.MapControllerRoute(
-    name: "default",
-    pattern: "{controller=Home}/{action=Index}/{id?}");
+app.UseEndpoints(endpoints =>
+{
+    endpoints.MapDefaultControllerRoute();
+    endpoints.MapReverseProxy();
+});
```
davidfowl, line 33:
```suggestion
app.MapDefaultControllerRoute();
app.MapReverseProxy();
```

twsouthwick: I thought the `UseEndpoints` would be better to configure metadata, etc. When do you recommend using UseEndpoints vs doing it from IAppBuilder?

src/SystemWebAdapters/src/Adapters/MigrationWebApplicationOptions.cs
twsouthwick: Yeah, very. I'm not sure if we actually need it after playing around with the VSIX tooling. It seems to work just fine without setting it, so I'll hold off on merging this in for now

src/SystemWebAdapters/samples/MvcCoreApp/Program.cs
davidfowl: WebApplication merges the concepts, there are advanced cases where you want to call UseEndpoints (like adding middleware after route execution) but this example doesn't need that AFAIK

twsouthwick:
This is no longer needed.

PR #467, opened 2022-03-29
Implement HttpCookieCollection and HttpCookie

twsouthwick:


src/SystemWebAdapters/src/HttpCookieCollection.cs
```diff
@@ -2,29 +2,122 @@
...
+        public ResponseCookies(IResponseCookies cookies)
+        {
+            _cookies = cookies;
+        }
+
+        public HttpCookie? this[string name] => throw new PlatformNotSupportedException("Getting response cookies by name is not supported");
+
+        public string[] Keys => throw new PlatformNotSupportedException("Keys are not available");
+
+        public int Count => throw new PlatformNotSupportedException("Count is not supported");
```
Tratcher, line 76: These can be implemented by parsing the Set-Cookie header, if needed.
https://github.com/dotnet/aspnetcore/blob/ea2b6d0af43aeaba39243d59d96eedd85089a1a2/src/Http/Headers/src/SetCookieHeaderValue.cs#L442

twsouthwick:
After looking at framework's implementation of HttpCookieCollection again, it's actually just a collection that gets written at the end. I've updated this to do the same by hooking into the HttpResponse.OnStarting event.

I've also added a couple of endpoints that manipulate cookies and they do it the same now on framework and core.

This will not keep the worlds in-sync (i.e. cookies set by using the adapters with the cookies set using IResponseCookies). However, this is similar to ASP.NET framework where if you're working directly with the header, things wouldn't show up there either.

Tratcher (approved):

PR #484, opened 2022-04-08
Add session sharing to System.Web.Adapters

mjrousos:
This updates System.Web.Adapter's HttpSessionState to allow both reading and writing session items from the remote app.


src/SystemWebAdapters/samples/ClassLibrary/SessionUtils.cs
```diff
@@ -8,5 +8,7 @@ public static void RegisterSessionKeys(RemoteAppSessionStateOptions options)
     {
         options.ApiKey = "test-key";
         options.RegisterKey<int>("test-value");
+        options.RegisterKey<int>("IntSessionItem");
```
twsouthwick, line 11: I see you using this for `SessionDemoModel`. What about storing that object in the session? It'll showcase that objects work too

src/SystemWebAdapters/samples/MvcCoreApp/MvcCoreApp.csproj
```diff
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Experimental.AspNetCore.Migrations" Version="1.0.0-preview-0001" />
```
twsouthwick, line 10: this isn't needed

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
```diff
@@ -1,121 +1,210 @@
...
+using System.Web.SessionState;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Net.Http.Headers;
 
 namespace System.Web.Adapters.SessionState;
 
-internal class RemoteAppSessionStateManager : ISessionManager
+internal class RemoteAppSessionStateManager : ISessionManager, IDisposable
```
twsouthwick, line 18: I'd prefer to keep the manager stateless and have the session state be the thing that knows how to commit itself/dispose/etc

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
```diff
@@ -1,121 +1,210 @@
...
-        if (!metadata.IsReadOnly)
+        if (_session is null)
         {
-            throw new NotSupportedException("Only readonly session is currently supported");
+            if (_contextAccessor.HttpContext is null)
+            {
+                throw new InvalidOperationException("Loading remote session state requires a current ASP.NET Core HttpContext");
+            }
+
+            using var timeout = new CancellationTokenSource(_options.CurrentValue.IOTimeoutSeconds * 1000);
```
twsouthwick, line 53: Could you use TimeSpan? I find it easier to read

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
```diff
@@ -1,121 +1,210 @@
...
-        _serializer = serializer;
+        _contextAccessor = contextAccessor ?? throw new ArgumentNullException(nameof(contextAccessor));
+        _options = options ?? throw new ArgumentNullException(nameof(options));
+        _httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
+        _serializer = serializer ?? throw new ArgumentNullException(nameof(serializer));
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
     }
 
-    public async Task<ISessionState> CreateAsync(HttpContextCore context, ISessionMetadata metadata)
+    public async Task<HttpSessionState> LoadAsync(bool readOnly)
```
twsouthwick, line 44: The point of passing the context in is so you don't need to access it via HttpContextAccessor which has its own perf issues

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
twsouthwick, line 53: Is this IOTimeoutSeconds something that should be configurable per endpoint?

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
```diff
@@ -1,121 +1,210 @@
...
-            throw new NotSupportedException("Only readonly session is currently supported");
+            if (_contextAccessor.HttpContext is null)
+            {
+                throw new InvalidOperationException("Loading remote session state requires a current ASP.NET Core HttpContext");
+            }
+
+            using var timeout = new CancellationTokenSource(_options.CurrentValue.IOTimeoutSeconds * 1000);
+            using var cts = CancellationTokenSource.CreateLinkedTokenSource(timeout.Token, _contextAccessor.HttpContext.RequestAborted);
+
+            // Ensure that session state is only loaded once, even if multiple threads attempt to access session state simultaneously.
```
twsouthwick, line 56: I don't think we need to track this - that's a bug in the code if they're using multiple threads to access it simultaneously. There's already locking being done on the server

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
```diff
@@ -1,121 +1,210 @@
...
-    private static HttpRequestMessage? PrepareHttpRequestMessage(HttpContextCore context, RemoteAppSessionStateOptions options, ISessionMetadata metadata)
+    public void Dispose()
     {
-        var cookie = GetAspNetCookie(context, options);
-
-        if (cookie is null && metadata.IsReadOnly)
-        {
-            return null;
-        }
+        _responseMessage?.Dispose();
```
twsouthwick, line 148: isn't this disposed in on line 141?

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
```diff
@@ -1,121 +1,210 @@
...
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(timeout.Token, _contextAccessor.HttpContext.RequestAborted);
 
-        using var stream = await response.Content.ReadAsStreamAsync();
+        try
+        {
+            // Mark the session complete so that no additional changes can be made
+            _session.Complete();
 
-        return await _serializer.DeserializeAsync(stream) ?? throw new InvalidOperationException("No session was found.");
+            using var request = await PrepareWriteRequestAsync(_session, cts.Token).ConfigureAwait(false);
```
twsouthwick, line 129: no need for the configure await in context of ASP.NET

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateOptions.cs
```diff
@@ -55,4 +55,12 @@ public class RemoteAppSessionStateOptions
     /// <param name="key"></param>
     public void RegisterKey<T>(string key) => KnownKeys.Add(key, typeof(T));
 
+#if NETCOREAPP3_1_OR_GREATER
+    /// <summary>
+    /// The maximum number of seconds loading session state from the remote app
+    /// or committing changes to it can take before timing out.
+    /// </summary>
+    [Required]
+    public int IOTimeoutSeconds { get; set; } = 60;
```
twsouthwick, line 64: name seems a little weird. Maybe NetworkTimeout? I believe the options binding works with Timespan which would negate the need for assuming a unit.

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateHandler.cs
```diff
@@ -37,12 +43,140 @@ public override async Task ProcessRequestAsync(HttpContext context)
         }
         else
         {
-            await Serializer.SerializeAsync(context.Session, context.Response.OutputStream, context.Request.TimedOutToken);
+            var readOnly = string.Equals(context.Request.Headers.Get(RemoteAppSessionStateOptions.ReadOnlyHeaderName), true.ToString(), StringComparison.OrdinalIgnoreCase);
```
twsouthwick, line 46:
```suggestion
            var readOnly = bool.TryParse(context.Request.Headers.Get(RemoteAppSessionStateOptions.ReadOnlyHeaderName), out var result) ? result : true;
```

twsouthwick: Makes it clear what you're doing and what the default value is (I'm assuming we should default to readonly if not set)

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateHandler.cs
```diff
@@ -37,12 +43,140 @@ public override async Task ProcessRequestAsync(HttpContext context)
...
+        {
+            session[key] = updatedSessionState[key];
+        }
+
+        foreach (var removedItem in updatedSessionState.RemovedKeys)
+        {
+            session.Remove(removedItem);
+        }
+
+        if (updatedSessionState.Abandon)
```
twsouthwick, line 172: This can be done first and then exit. Other changes don't matter if session is abandoned

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateModule.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        application.PostMapRequestHandler += Application_PostMapRequestHandler;
+    }
+
+    private void Application_PostMapRequestHandler(object sender, EventArgs e)
+    {
+        var context = ((HttpApplication)sender).Context;
+
+        // Check whether the request is for the session state path
+        if (context.Request.AppRelativeCurrentExecutionFilePath.Equals("~/session-state", StringComparison.OrdinalIgnoreCase))
```
twsouthwick, line 26: the module system I set up has some options that includes a path

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateModule.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            // Set the handler before setting session state behavior since
+            // setting the handler can override session state behavior
+            context.Handler = new RemoteAppSessionStateHandler();
```
twsouthwick, line 30: there's no need to create this each time. the previous module was setting the event as a local delegate so that it could capture the single instance of the handler

src/SystemWebAdapters/src/SessionState/HttpSessionState.cs
```diff
@@ -1,46 +1,135 @@
...
 
+using System.Collections;
+using System.Linq;
 using System.Web.Adapters;
+using System.Web.Adapters.SessionState;
 
 namespace System.Web.SessionState;
 
 public class HttpSessionState
+    // Implement ICollection for compatibility with .NET Framework
```
twsouthwick, line 12: Do we need this? If so you need to update the ref file with `./update-apis.ps1`

src/SystemWebAdapters/src/SessionState/HttpSessionState.cs
```diff
@@ -1,46 +1,135 @@
...
+    private bool _completed;
 
-    public HttpSessionState(ISessionState container)
+    internal bool Completed => _completed;
+
+    internal ISessionUpdate Updates => _stateUpdate;
+
+    internal void Complete() => _completed = true;
+
+    public HttpSessionState(ISessionState remoteState)
```
twsouthwick, line 25: Any methods we need to track here should be on ISessionState. That way anyone can implement it how they need.

twsouthwick, line 25: nit: It's not all remote state

twsouthwick (commented):
Can you set up some tests for the module and middleware?

src/SystemWebAdapters/src/Adapters/ISessionManager.cs
```diff
@@ -2,18 +2,21 @@
...
-public interface ISessionManager
+public interface ISessionManager : IDisposable
 {
     /// <summary>
     /// Creates an instance of <see cref="ISessionState"/> for a given content and metadata.
     /// </summary>
     /// <param name="context">Current <see cref="HttpContextCore"/>.</param>
     /// <param name="metadata">Metadata for the session.</param>
-    Task<ISessionState> CreateAsync(HttpContextCore context, ISessionMetadata metadata);
+    Task<HttpSessionState> LoadAsync(bool readOnly);
```
twsouthwick, line 19: Why remove the context?

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateHandler.cs
```diff
@@ -1,17 +1,23 @@
...
 
-public sealed class RemoteAppSessionStateHandler : HttpTaskAsyncHandler, IRequiresSessionState, IReadOnlySessionState
+public sealed class RemoteAppSessionStateHandler : HttpTaskAsyncHandler
 {
     private SessionSerializer? _serializer;
 
     private static readonly RemoteAppSessionStateOptions _options = new();
 
+    // Track locked sessions awaiting updates or release
+    private static readonly ConcurrentDictionary<string, Channel<ISessionUpdate?>> SessionDataChannels = new();
```
twsouthwick, line 19: Can you move this concept of tracking to its own class?

src/SystemWebAdapters/samples/MvcCoreApp/MvcCoreApp.csproj
mjrousos: I think I have an old version of the YARP proxy stuff. I'll remove this.

src/SystemWebAdapters/samples/ClassLibrary/SessionUtils.cs
mjrousos: Good idea. I'll do that!

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
mjrousos: Sure, I'll switch.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateOptions.cs
mjrousos: Good suggestion. I'll switch to Timespan and rename this.

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateHandler.cs
mjrousos: I'm happy to reword this to clarify, but I was thinking the default would be read/write. It probably doesn't really matter since we're the only caller of this endpoint and can be explicit, but if the user doesn't specify it seemed like we should assume they may want to make changes.

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateModule.cs
mjrousos: Good point. I'll do that.

src/SystemWebAdapters/src/SessionState/HttpSessionState.cs
mjrousos: We don't. I just thought there may be end-users who depend on the fact that HttpSessionState implements ICollection in NetFx. I'll update the ref file.

mjrousos:
Thanks for the review, @twsouthwick. Yea, adding some tests is on my todo list.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
mjrousos: That makes sense. I'll go back to passing the context.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
mjrousos: I don't think so. I'd think the time to load (all) session state or commit changes wouldn't vary between endpoints (except that endpoints with session disabled won't use this at all).

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
mjrousos: If we somehow (likely due to a bug) end up disposing the manager without calling CommitAsync, this makes sure that we still dispose the response.

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateHandler.cs
mjrousos: Good point. Re-ordered.

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateModule.cs
mjrousos: Using that now, post-merge

src/SystemWebAdapters/src/Adapters/ISessionManager.cs
mjrousos: It didn't seem necessary since the manager could just get it on its own. You brought up a good point about HttpContextAccessor perf in another comment, though, so I've added it back.

src/SystemWebAdapters/src/SessionState/HttpSessionState.cs
mjrousos: It is now :) With my changes, this will be the initial state snapshot (which comes from the remote machine) and updates are stored in a different object so that it's easy to commit just the diff of session state when we're done with the session.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
mjrousos: I just always include ConfigureAwait(false) if I'm in a library. I suppose that chances of this being used outside of an ASP.NET scenario is pretty slim.  Technically this is a library that _could_ be used from elsewhere to load remote session state. 

Probably the dependency on `Microsoft.AspNetCore.Http.HttpContext` means that won't happen, though. I'll remove the ConfigureAwaits.

src/SystemWebAdapters/src/Ref.Standard.cs
```diff
@@ -290,6 +287,17 @@ public enum SameSiteMode
         Strict = 2,
     }
 }
+namespace System.Web.Adapters
+{
+    public partial interface ISessionUpdate
```
twsouthwick, line 292: The adapter interfaces are currently not exposed on anything but .NET Core. Is there a reason to do so here?

src/SystemWebAdapters/src/Ref.Standard.cs
```diff
@@ -305,18 +313,23 @@ public partial class Cache
 }
 namespace System.Web.SessionState
 {
-    public partial class HttpSessionState
+    public partial class HttpSessionState : System.Collections.ICollection, System.Collections.IEnumerable
     {
         internal HttpSessionState() { }
         public int Count { get { throw new System.PlatformNotSupportedException("Only support when running on ASP.NET Core or System.Web");} }
-        public bool IsNewSession { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw new System.PlatformNotSupportedException("Only support when running on ASP.NET Core or System.Web");} }
+        public bool HasUpdates { get { throw new System.PlatformNotSupportedException("Only support when running on ASP.NET Core or System.Web");} }
```
twsouthwick, line 320: You cannot add methods here that do not exist on framework

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateHandler.Framework.cs
```diff
@@ -27,12 +33,141 @@ public override async Task ProcessRequestAsync(HttpContext context)
...
+                context.Response.StatusCode = 200;
+                await _serializer.SerializeAsync(context.Session, context.Response.OutputStream, cts.Token).ConfigureAwait(false);
+
+                // Delimit the json body with a new line to mark the end of content
+                context.Response.Write('\n');
+                await context.Response.FlushAsync().ConfigureAwait(false);
+
+                // Wait for up to request timeout for updated session state to be provided
+                // (or for null to be passed as updated session state to indicate no updates).
+                var updatedSessionState = await responseChannel.Reader.ReadAsync(cts.Token).ConfigureAwait(false);
```
twsouthwick, line 88: I'm not seeing why we need this to be a channel. Why not use a `TaskCompletionSource`?

src/SystemWebAdapters/src/SessionState/HttpSessionState.cs
twsouthwick: But not HttpSessionState - that is *any* state object. It does not know about what kind of session state it is

src/SystemWebAdapters/src/Ref.Standard.cs
```diff
@@ -1,6 +1,3 @@
-// Licensed to the .NET Foundation under one or more agreements.
```
twsouthwick, line 1: This needs to be recopied - doesn't get added by the tool we use

src/SystemWebAdapters/src/SessionState/HttpSessionState.cs
twsouthwick: Is there a reason to put the comment on there? I haven't called out other APIs that are there for .NET Framework compat (that dictates most of the shape here)

src/SystemWebAdapters/src/Ref.Standard.cs
```diff
@@ -305,18 +313,23 @@ public partial class Cache
 }
 namespace System.Web.SessionState
 {
-    public partial class HttpSessionState
+    public partial class HttpSessionState : System.Collections.ICollection, System.Collections.IEnumerable
```
twsouthwick, line 316: If we're going to add these, then ISessionState should have them to so it can be delegate through to that object.

src/SystemWebAdapters/src/SessionState/HttpSessionState.cs
mjrousos: That makes sense. I'll re-work it like that and see how it turns out.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
mjrousos: Manager is now stateless and the session state commits/disposes itself.

src/SystemWebAdapters/src/SessionState/HttpSessionState.cs
mjrousos: Done

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateHandler.cs
mjrousos: Good idea. I'll see what I can do.

src/SystemWebAdapters/src/Ref.Standard.cs
mjrousos: Originally it was because this interface was used by the session handler, too, but that's no longer the case so these can now be .NET-only.

src/SystemWebAdapters/src/Ref.Standard.cs
mjrousos: Good catch. Fixed.

src/SystemWebAdapters/src/Ref.Standard.cs
mjrousos: Updated

src/SystemWebAdapters/src/Ref.Standard.cs
mjrousos: Good catch. Fixed.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateHandler.Framework.cs
mjrousos: That's a good point. For some reason my mind just always goes to channels for asynchronous object production/passing. In this case we're only waiting on a single item so a `TaskCompletionSource` would probably be fine (and simpler). I'll try that out and see if there's any advantage to the channel or if I should just switch.

src/SystemWebAdapters/src/SessionState/HttpSessionState.cs
mjrousos: Fixed

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
mjrousos: I agree this shouldn't matter in practice (barring a bug in how we use the manager) but I like the idea of preventing unnecessary HTTP requests.

src/SystemWebAdapters/src/Adapters/ISessionState.cs
```diff
@@ -1,32 +1,29 @@
...
     bool IsNewSession { get; }
 
     void Abandon();
 
-    object? this[string name] { get; set; }
-
-    void Add(string name, object value);
-
-    void Remove(string name);
+    Task LoadAsync(HttpContextCore context, bool readOnly, CancellationToken cancellationToken = default);
```
twsouthwick, line 26: Please remove. It's expected that this is loaded and available when you have the ISessionState. It appears you're only using it internally anyway

src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
```diff
@@ -31,10 +31,21 @@ private async Task ManageStateAsync(HttpContextCore context, ISessionMetadata me
...
-
-        await _next(context);
+        using var state = await manager.CreateAsync(context, metadata.IsReadOnly);
+
+        try
+        {
+            context.Features.Set(new HttpSessionState(state));
+            await _next(context);
+        }
+        finally
```
twsouthwick, line 41: should this be in a finally? do we want to commit if there's a failure.

src/SystemWebAdapters/src/Adapters/ISessionState.cs
```diff
@@ -1,32 +1,29 @@
...
 using System.Web.SessionState;
 
 namespace System.Web.Adapters;
 
 /// <summary>
-/// Represents the state of a session and is used to create a <see cref="HttpSessionState"/> . Disposing the state will handle any writing that may need to be done.
+/// Represents the state of a session and is used to create a <see cref="HttpSessionState"/>.
 /// </summary>
-public interface ISessionState : IAsyncDisposable
+public interface ISessionState : IDictionary<string, object?>, IDisposable
```
twsouthwick, line 14: Why change to IDisposable from IAsyncDisposable?

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionDataResponse.cs
```diff
@@ -0,0 +1,15 @@
+using System.Net.Http;
+
+namespace System.Web.Adapters.SessionState;
+
+internal class RemoteSessionDataResponse
```
twsouthwick, line 5: this should implement IDisposable to manage the HttpResponse

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionState.cs
```diff
@@ -0,0 +1,296 @@
...
+    private RemoteSessionData? _remoteData;
+
+    public RemoteSessionState(RemoteSessionService remoteSessionService, RemoteAppSessionStateOptions options, ILogger<RemoteSessionState> logger)
+    {
+        _remoteSessionService = remoteSessionService ?? throw new ArgumentNullException(nameof(remoteSessionService));
+        _options = options ?? throw new ArgumentNullException(nameof(options));
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    public async Task LoadAsync(HttpContextCore context, bool readOnly, CancellationToken cancellationToken = default)
```
twsouthwick, line 32: Could this be a static factory to load it (assuming it's being done this way due to no async constructors). I'd prefer we don't need to check if things are available so much in the class and just initialize everything initially

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionState.cs
```diff
@@ -0,0 +1,296 @@
...
+        if (!_loaded)
+        {
+            using var timeout = new CancellationTokenSource(_options.NetworkTimeout);
+            using var cts = CancellationTokenSource.CreateLinkedTokenSource(timeout.Token, context.RequestAborted, cancellationToken);
+
+            // Ensure that session state is only loaded once, even if multiple threads attempt to access session state simultaneously.
+            // Even though ASP.NET Core request handling is usually expected to be single-threaded, it's good to guarantee only one thread
+            // loads session data since if two threads were to make the call one would end up blocked waiting for the session state to
+            // unlock.
+            await _remoteLoadSemaphore.WaitAsync(cts.Token);
```
twsouthwick, line 43: This is unnecessary - ASP.NET Core does not do checks like this, so we don't need to

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionState.cs
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        using var timeout = new CancellationTokenSource(_options.NetworkTimeout);
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(timeout.Token, context.RequestAborted, cancellationToken);
+
+        var sessionId = _remoteData.SessionID;
+
+        try
+        {
+            // Note that this is not thread safe and callers should not be writing session state while committing
```
twsouthwick, line 95: that is the case for everything in ASP.NET Core :)

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionState.cs
```diff
@@ -0,0 +1,296 @@
...
+            // when session data is loaded.
+            var sessionId = context.Request.Cookies[_options.CookieName];
+
+            try
+            {
+                if (!_loaded)
+                {
+                    // Get or create session data
+                    var response = await _remoteSessionService.GetSessionDataAsync(sessionId, readOnly, context, cts.Token);
+                    _remoteData = response.RemoteSessionData;
```
twsouthwick, line 56: why not just store this remote object?

src/SystemWebAdapters/src/Adapters/ISessionState.cs
twsouthwick, line 14: This adds a number of methods to implement that are not needed for HttpSessionState. Why not just `ICollection`?

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateHandler.Framework.cs
```diff
@@ -27,12 +33,137 @@ public override async Task ProcessRequestAsync(HttpContext context)
...
+            context.Response.StatusDescription = "No session ID found";
+            context.Response.End();
+            return;
+        }
+
+        // Get the channel that will be used to write the updated session data
+        // to the in-progress request that will update session data
+        if (SessionDataChannels.TryGetValue(sessionId, out var channel))
+        {
+            if (channel.Writer.TryWrite(sessionData))
```
twsouthwick, line 129: Do you want to eject the channel from the dictionary at this point? Is it needed after this?

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateHandler.Framework.cs
mjrousos: Fixed.

src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
mjrousos: Yes, otherwise we don't release the lock.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionState.cs
mjrousos: Yes, I'll make that change. I set it up this way initially to allow for loading to happen later (in lazy scenarios if we enabled them). The way you have lazy loading setup makes this unnecessary, though, so I'll fix it up to create via am async factory method.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateManager.cs
mjrousos: With this changing to a factory method instead of a load method, this isn't needed anymore. I think it's possible we'll still need to synchronize this in the future if there are other components that need to load RemoteSessionState (maybe an ASP.NET Core ISession implementation?) but the synchronization wouldn't happen here anymore, so I'll remove this for now and we can worry about that later.

src/SystemWebAdapters/src/Adapters/ISessionState.cs
mjrousos: Fixed

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionDataResponse.cs
mjrousos: Fixed

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionState.cs
mjrousos: Removed

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionState.cs
mjrousos: Good point. Removed.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionState.cs
mjrousos: Fixed

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateHandler.Framework.cs
mjrousos: Removing the channel from the dictionary is already handled in `GetSessionStateAsync` (which should happen as soon as the session data is written here) so it isn't necessary here.

src/SystemWebAdapters/src/Adapters/ISessionState.cs
mjrousos: The only disposal needed is for the HTTP response so there's nothing async to do.

mjrousos: Changed

src/SystemWebAdapters/src/Adapters/SessionState/RemoteAppSessionStateHandler.Framework.cs
```diff
@@ -27,12 +33,137 @@ public override async Task ProcessRequestAsync(HttpContext context)
...
+
+
+    private async Task GetSessionStateAsync(HttpContext context, bool exclusive)
+    {
+        var timeoutCts = new CancellationTokenSource(context.Server.ScriptTimeout * 1000);
+        var cts = CancellationTokenSource.CreateLinkedTokenSource(timeoutCts.Token, context.Response.ClientDisconnectedToken);
+
+        if (exclusive)
+        {
+            // If session data is retrieved exclusively, then it needs sent to the client and
```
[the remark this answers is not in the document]
mjrousos: Sure. I've refactored it a bit to do that.

src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
```diff
@@ -45,11 +45,14 @@ private async Task ManageStateAsync(HttpContextCore context, ISessionMetadata me
...
             await _next(context);
-            await state.CommitAsync(context.RequestAborted);
         }
         finally
         {
             context.Features.Set<HttpSessionState?>(null);
+
+            // Commit in the finally block so that session doesn't remain locked if
+            // an exception is thrown
+            await state.CommitAsync(context.RequestAborted);
```
twsouthwick, line 55: Do we want to commit at this point if an error has occurred?

src/SystemWebAdapters/src/Framework/RemoteAppSessionStateHandler.cs
mjrousos: Looking at this more, I could have a helper class that wraps TaskCompletionSources so that this is abstracted some and the handler would just interact with that class, but I don't think this is complex enough to need another class to separate it out.

Let me know if you feel strongly about this and I can refactor it some.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionService.cs
```diff
@@ -0,0 +1,99 @@
...
+    private readonly RemoteAppSessionStateOptions _options;
+
+    public RemoteSessionService(HttpClient client, ILogger<RemoteSessionService> logger, SessionSerializer serializer, IOptions<RemoteAppSessionStateOptions> options)
+    {
+        _client = client ?? throw new ArgumentNullException(nameof(client));
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
+        _serializer = serializer ?? throw new ArgumentNullException(nameof(serializer));
+        _client.BaseAddress = new Uri(_options.RemoteApp, _options.SessionEndpointPath);
+        _client.DefaultRequestHeaders.Add(_options.ApiKeyHeader, _options.ApiKey);
```
twsouthwick, line 27: This is fine, but we'll want to see if there's a need to react to ApiKey changes without restarting the service

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionService.cs
```diff
@@ -0,0 +1,99 @@
...
+        AddSessionCookieToHeader(req, sessionId);
+        AddReadOnlyHeader(req, readOnly);
+
+        var response = await _client.SendAsync(req, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+        _logger.LogTrace("Received {StatusCode} response getting remote session state for request {RequestUrl}", response.StatusCode, callingContext.Request.GetDisplayUrl());
+        response.EnsureSuccessStatusCode();
+
+        // Only read until the first new line since the response is expected to remain open until
+        // RemoteAppSessionStateManager.CommitAsync is called.
+        using var streamReader = new StreamReader(await response.Content.ReadAsStreamAsync());
```
twsouthwick, line 42: This feels somewhat fragile (can a newline show up in properly format json?). What about encoding a length up front that we then know how much to read or having a sentinel marker at the end to know the "frame" is done?

In any case, can you move it to a separate method for the logical act of "reading a frame"?

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionState.cs
```diff
@@ -0,0 +1,161 @@
...
+
+    public bool IsSynchronized => false;
+
+    public object SyncRoot => this;
+
+    public void Abandon() => RemoteData.Abandon = true;
+
+    public void Add(string key, object? value) => RemoteData.Values[key] = value;
+
+    public void Add(KeyValuePair<string, object?> item) => Add(item.Key, item.Value);
```
twsouthwick, line 118: I think a number of these are not needed anymore

src/SystemWebAdapters/src/Adapters/SystemWebAdaptersExtensions.cs
```diff
@@ -23,7 +23,7 @@ public static ISystemWebAdapterBuilder AddSystemWebAdapters(this IServiceCollect
         public static ISystemWebAdapterBuilder AddSessionManager<TManager>(this ISystemWebAdapterBuilder builder)
             where TManager : class, ISessionManager
         {
-            builder.Services.AddSingleton<ISessionManager, TManager>();
+            builder.Services.AddTransient<ISessionManager, TManager>();
```
twsouthwick, line 26: is this change needed still?

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionData.cs
```diff
@@ -0,0 +1,19 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Web.Adapters.SessionState;
+
+internal class RemoteSessionData
+{
+    public bool Abandon { get; set; }
```
twsouthwick, line 8:
```suggestion
    public bool IsAbandoned { get; set; }
```

twsouthwick (approved):
A couple of comments, but otherwise LGTM

src/SystemWebAdapters/src/Adapters/ISessionManager.cs
```diff
@@ -11,9 +11,9 @@ namespace System.Web.Adapters;
 public interface ISessionManager
 {
     /// <summary>
-    /// Creates an instance of <see cref="ISessionState"/> for a given content and metadata.
+    /// Creates an instance of <see cref="ISessionState"/> for a given context.
     /// </summary>
     /// <param name="context">Current <see cref="HttpContextCore"/>.</param>
-    /// <param name="metadata">Metadata for the session.</param>
+    /// <param name="readOnly">Whether the session should be created read-only.</param>
```
twsouthwick, line 17: this rewording doesn't apply now

src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
mjrousos: I think we do because otherwise the session remains locked. I suppose it's possible that we'll end up committing partial updates to session items if something fails in the middle of a component working with session state, but I think that's not super likely (and not submitting changes when some were previously made could also be undesirable in some scenarios).

I think it makes sense to leave this as is so we don't end up with abandoned locked sessions and if we hear feedback that committing changes after an error is problematic, we can change this to release the lock without applying changes at that point.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionService.cs
mjrousos: I tested for \n support and it works (System.Text.Json doesn't insert \n and any \n in serialized strings are escaped). I'll pull this out into its own method, though, for cleaner logical separation.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionState.cs
mjrousos: Good catch. I removed the ones that aren't part of ISessionState

src/SystemWebAdapters/src/Adapters/SystemWebAdaptersExtensions.cs
mjrousos: I think it's still better as transient because the manager has a dependency on RemoteSessionService which is a typed HttpClient and it makes sense to let the HttpClientFactory manage the lifetime of HttpClients.

src/SystemWebAdapters/src/Adapters/SessionState/RemoteSessionData.cs
mjrousos: Fixed.

src/SystemWebAdapters/src/Adapters/ISessionManager.cs
mjrousos: Good catch. Fixed.

src/SystemWebAdapters/src/Adapters/SystemWebAdaptersExtensions.cs
twsouthwick: I thought that you could cache the HttpClient itself and the message handler underneath handles stuff.

In any case, maybe then just remove this helper method. If it doesn't make sense for all cases, we probably don't need it

src/SystemWebAdapters/src/Adapters/ISessionManager.cs
```diff
@@ -11,9 +11,9 @@ namespace System.Web.Adapters;
 public interface ISessionManager
 {
     /// <summary>
-    /// Creates an instance of <see cref="ISessionState"/> for a given content and metadata.
+    /// Creates an instance of <see cref="ISessionState"/> for a given context.
     /// </summary>
     /// <param name="context">Current <see cref="HttpContextCore"/>.</param>
-    /// <param name="metadata">Metadata for the session.</param>
+    /// <param name = "metadata" > Metadata for the session.</param>
```
twsouthwick, line 17:
```suggestion
    /// <param name="metadata" > Metadata for the session.</param>
```

src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
twsouthwick: My question is more if things have failed, do we still commit (in general, not specifically to handle locked state)? I don't believe that happens on framework (we can verify).

If the issue is that on the framework server we need to unlock, isn't that why we keep a connection open? Doesn't that get closed in the IDispose? Any removal of session locking can occur at the disconnection on ther framework server.

twsouthwick: looking at your implementation, wouldn't a OperationCanceledException be thrown and then the TaskCompletionSource be cleaned up and the session managed?

twsouthwick: fyi: https://referencesource.microsoft.com/#System.Web/State/SessionStateModule.cs,1337

Looks like if there was an error, the session is not updated

src/SystemWebAdapters/src/Adapters/Modules/RemoteSessionModule.Framework.cs
```diff
@@ -27,13 +27,42 @@ void MapRemoteSessionHandler(object sender, EventArgs e)
...
+                // Getting the session with non-exclusive access requires read-only session
+                else if (!exclusive && httpMethod.Equals("GET", StringComparison.OrdinalIgnoreCase))
+                {
+                    context.SetSessionStateBehavior(SessionStateBehavior.ReadOnly);
+                }
+
+                // Setting session with exclusive access requires NO session because the session
+                // should have already been previously locked by an exclusive get call.
+                // PUT calls should typically be exclusive, but callers might pass !exclusive
+                // if there are no updates being made and the call is only made to release a lock
```
twsouthwick, line 52: readonly session shouldn't be locked by the handler here as ASP.NET isn't locking it

mjrousos: Yea, this comment is wrong. I'll update this comment. In the case of a PUT call, we're releasing/updating an existing exclusive session access so the exclusive/read-only parameter doesn't really matter.

src/SystemWebAdapters/src/Adapters/SystemWebAdaptersExtensions.cs
mjrousos: We still need to register an ISessionManager, but I agree the helper method isn't needed since it's only making that one call which `AddRemoteAppSession` can do directly.

src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
mjrousos: I'll have to test to see how long it takes between the client disconnecting and the server firing ClientDisconnectedToken. I'm not sure we make any guarantees about that happening quickly, so I was thinking of the ClientDisconnectedToken as a backup plan, not an expected flow. I can look into it.

mjrousos: No calling CommitAsync in failing cases *should* work but I'd like to double-check how long it takes between closing the HttpResponse (which should happen when leaving `ManageStateAsync`) and the `ClientDisconnectedToken` token firing in the handler on the server. I'm not sure how long that takes (especially since we're not even cancelling the request, just closing the response; not sure if that makes a difference or not). If it's pretty quick, then we don't need to call CommitAsync in failing cases.

twsouthwick: If you do need to call the put for that, let's add IAsyncDisposable back and do it there to keep the semantics the same.

mjrousos: Hm... for some reason in my testing, I don't see `ClientDisconnectedToken` firing even after disposing of the response on the client side. I'm not sure if I'm doing something wrong or if that feature is just unreliable (other folks have found it [inconsistent](https://stackoverflow.com/questions/41487714/c-sharp-webmethod-does-not-receive-clientdisconnectedtoken-on-azure) in the past).

mjrousos: Given the issues with ClientDisconnectedToken, I've re-worked this to implement IAsyncDisposable and explicitly release the session lock (without making updates) in the dispose path.

src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
```diff
@@ -45,14 +45,16 @@ private async Task ManageStateAsync(HttpContextCore context, ISessionMetadata me
         try
         {
             await _next(context);
+
+            // Commit changes to session state and release the lock
```
twsouthwick, line 49: This is an implementation detail of the remote session

mjrousos: Good point. I'll remove this comment (and maybe put something to the same effect in RemoteSessionState)

PR #486, opened 2022-04-11
Implement HttpServerUtility

twsouthwick:
This implements HttpServerUtility except MapPath which ends up bringing in a number of other things. This will be done in a separate PR.

### DIFF
--- a/src/SystemWebAdapters/src/HttpServerUtility.cs
+++ b/src/SystemWebAdapters/src/HttpServerUtility.cs
@@ -16,7 +16,7 @@ public class HttpServerUtility
 
     public string MachineName => Environment.MachineName;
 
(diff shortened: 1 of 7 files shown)


src/SystemWebAdapters/src/HttpServerUtility.cs
```diff
@@ -1,27 +1,117 @@
...
+    private readonly HttpContextCore _context;
+
+    internal HttpServerUtility(HttpContextCore context)
+    {
+        _context = context;
+    }
+
+    public string MachineName => Environment.MachineName;
+
+    public string MapPath(string path) => throw new NotImplementedException();
```
davidfowl, line 19: This will be interesting...

Tratcher (approved):

PR #488, opened 2022-04-13
Add proxy header support for adapters

twsouthwick:
This change adds a module for the framework app that will handle proxy values. As part of this change, the SystemWebAdapter module will load any registered modules so that we can keep them specific to the feature they enable.

### DIFF
--- a/src/SystemWebAdapters/src/Adapters/Modules/ProxyHeaderModule.Framework.cs
+++ b/src/SystemWebAdapters/src/Adapters/Modules/ProxyHeaderModule.Framework.cs
@@ -1,0 +1,94 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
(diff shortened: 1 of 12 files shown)


src/SystemWebAdapters/src/Adapters/Modules/ProxyHeaderModule.Framework.cs
```diff
@@ -0,0 +1,94 @@
...
+    private struct ProtoHost
+    {
+        public ProtoHost(string host)
+        {
+            var idx = host.IndexOf(":");
+
+            if (idx < 0)
+            {
+                ServerName = host;
+                Port = "80";
```
Tratcher, line 81: Better to default to 443 for HTTPS, or no port if possible.

Tratcher (approved):
We should confirm if the multi-site model will work for the tooling folks, otherwise there's no need for the host forwarders.

PR #489, opened 2022-04-14
Add SessionBehavior to enable loading session on demand

twsouthwick:
This adds an enum that allows for setting behavior for session to be none, eager loading (i.e. async in middleware), and on-demand that will only be when it is used.

### DIFF
--- a/src/SystemWebAdapters/src/Adapters/SessionBehavior.cs
+++ b/src/SystemWebAdapters/src/Adapters/SessionBehavior.cs
@@ -3,9 +3,11 @@
 
 namespace System.Web.Adapters;
 
-public interface ISessionMetadata
+public enum SessionBehavior
 {
-    SessionBehavior Behavior { get; }
+    None,
+    Eager,
 
-    bool IsReadOnly { get; }
+    [Obsolete("This will enable session on the endpoint but will resort to async over sync behavior")]
+    OnDemand,
 }
--- a/src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
+++ b/src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
@@ -21,7 +21,7 @@ internal class SessionMiddleware
     }
 
     public Task InvokeAsync(HttpContextCore context)
-        => context.GetEndpoint()?.Metadata.GetMetadata<ISessionMetadata>() is { IsEnabled: true } metadata
+        => context.GetEndpoint()?.Metadata.GetMetadata<ISessionMetadata>() is { Behavior: not SessionBehavior.None } metadata
             ? ManageStateAsync(context, metadata)
             : _next(context);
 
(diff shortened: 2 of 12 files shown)


src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
```diff
@@ -31,10 +31,49 @@ private async Task ManageStateAsync(HttpContextCore context, ISessionMetadata me
 
         var manager = context.RequestServices.GetRequiredService<ISessionManager>();
 
-        await using var state = await manager.CreateAsync(context, metadata);
+#pragma warning disable CS0618 // Type or member is obsolete
```
mjrousos, line 34: I think this might be simpler if we leave this unchanged and depend on the manager to do the right thing based on metadata.Behavior. The manager could create the delegating state if it wants to or, the way I imagined it, the managed could just return the session state implementation it would have anyhow but without calling LoadAsync (and let it happen as-needed in the session implementation).

mjrousos (approved):
Looks good. I was planning to update my PR to allow lazily loading within the RemoteSessionState implementation so we may not need the delegating state here - just let the manager do what it needs to based on the metadata.Behavior value.

src/SystemWebAdapters/src/Adapters/SessionMiddleware.cs
mjrousos: Thinking more about this, this way is probably just as good so I think it's fine as-is.

mjrousos (approved):
LGTM

twsouthwick:
@Tratcher do you have any concerns with allowing session to fallback to sync-over-async if opted into?

src/SystemWebAdapters/src/Adapters/SessionBehavior.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Web.Adapters;
+
+public enum SessionBehavior
+{
+    None,
+    Eager,
```
Tratcher, line 9:
```suggestion
    /// Asynchronously loads the session for controllers with this attribute before running the controller.
    PreLoad,
```

src/SystemWebAdapters/src/Adapters/SessionBehavior.cs
```diff
@@ -0,0 +1,13 @@
...
+
+    [Obsolete("This will enable session on the endpoint but will resort to async over sync behavior")]
```
Tratcher, line 11:
```suggestion
    [Obsolete("This will enable session on the endpoint but will resort to sync over async behavior")]
```

Tratcher (approved):

PR #490, opened 2022-04-16
Rename to SessionBehavior.PreLoad

twsouthwick:


twsouthwick:
Missed a couple of recommendations in https://github.com/aspnet/AspLabs/pull/489

PR #495, opened 2022-04-19
Separate serialization from remote app state

twsouthwick:
This change provides simple roundtrip utilities to serialize/deserialize HttpSessionState and ISessionState objects. This will simplify the ability to try out other strategies for session management.


Tratcher (approved):

src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SessionState/Serialization/ISessionSerializer.Framework.cs
```diff
@@ -0,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Web.SessionState;
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.SessionState.Serialization;
+
+public interface ISessionSerializer
```
mjrousos, line 9: I like this abstraction; thanks for adding it.

src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SessionState/Serialization/ISessionSerializer.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.AspNetCore.SystemWebAdapters.SessionState.Serialization;
+
+public interface ISessionSerializer
+{
+    ISessionState? Deserialize(string? data);
```
mjrousos, line 8: For consistency's sake should this take a byte[]? I would imagine we can get that as easily as a string from the HTTP response and it would be a more flexible input for future ISessionSerializer implementations.

src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SessionState/Serialization/SerializedSessionState.cs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    [JsonPropertyName("id")]
+    public string SessionID { get; set; } = null!;
+
+    [JsonPropertyName("r")]
+    public bool IsReadOnly { get; set; }
+
+    [JsonIgnore]
+    public SessionValues Values => RawValues ??= new();
```
mjrousos, line 32: Why not just serialize this directly (as "v") instead of having both `Values` and `RawValues`?

src/SystemWebAdapters/test/Microsoft.AspNetCore.SystemWebAdapters.Tests/Adapters/SessionState/SessionStateSerialization.cs
```diff
@@ -1,155 +1,141 @@
...
-        using var stream = new MemoryStream(Encoding.UTF8.GetBytes(PayLoad));
-
         // Act
-        var result = await serializer.DeserializeAsync(stream);
+        var result = serializer.Deserialize(PayLoad);
 
         // Assert
         Assert.NotNull(result);
-        Assert.Empty(result!.Values);
+        Assert.Equal(0, result!.Count);
```
mjrousos, line 28: Nit-pick: Why not Assert.Empty(result)? I thought that was preferred over comparing to a collection's `Count` directly?

src/SystemWebAdapters/test/Microsoft.AspNetCore.SystemWebAdapters.Tests/Adapters/SessionState/SessionStateSerialization.cs
```diff
@@ -161,29 +147,25 @@ public async Task ComplexObject()
...
     }
 
-    private static void AssertValuesEqual((string, object?)[] expected, RemoteSessionData? result) =>
-    Assert.Collection(result!.Values.Keys, expected.Select<(string, object?), Action<string>>(expected =>
-                 actual =>
-                 {
-                     Assert.Equal(expected.Item1, actual);
-                     Assert.Equal(expected.Item2, result.Values[actual]);
-                 }).ToArray());
+    private static void AssertKey<T>(ISessionState state, string key, T expected)
```
mjrousos, line 159: Nit-pick: The old code here (AssertValues) had the added benefit of making sure the *number* of values in the state were right, too.

src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/ISessionState.cs
```diff
@@ -28,14 +28,12 @@ public interface ISessionState : IAsyncDisposable
 
     object SyncRoot { get; }
 
-    void Abandon();
+    bool IsAbandoned { get; set; }
```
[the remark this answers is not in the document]
twsouthwick: If someone needs more method style work there, they can do it since it's a property :) When I looked in the framework version, things would just use properties, so it made it simpler (and gave the ability to gather everything needed for serialization) to make that the contract we implement HttpSessionState against.

src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SessionState/RemoteSession/RemoteSessionState.cs
```diff
@@ -2,141 +2,55 @@
...
+            _response = null;
         }
-        catch (Exception exc)
+
+        if (_commitOrRelease is { } onCommit)
         {
-            logger.LogError(exc, "Unable to load remote session state for session {SessionId}", sessionId);
-            throw;
+            _commitOrRelease = null;
+            await onCommit(null, default);
```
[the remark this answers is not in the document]
twsouthwick: Updated

src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SessionState/Serialization/SerializedSessionState.cs
twsouthwick: The Values will always have a value, while RawValues may be null (i.e. no keys yet). This prevents us from creating the SessioNValues object if we don't have a need (it'll create it on demand)

src/SystemWebAdapters/test/Microsoft.AspNetCore.SystemWebAdapters.Tests/Adapters/SessionState/SessionStateSerialization.cs
twsouthwick: result doesn't implement IEnumerable now as it causes weirdness with serialization. I also thought it was weird that the enumeration was the keys - so now ISessionState exposes a `.Key` property and we create the enumeration on HttpSessionState from that. That being said, `Assert.Empty` doesn't work now on this type

mjrousos: That makes sense

src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/Adapters/SessionState/Serialization/SerializedSessionState.cs
mjrousos: Could Values just be nullable?

PR #496, opened 2022-04-20
Separate adapters session logic from core package

twsouthwick:


src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
```diff
@@ -46,16 +50,9 @@
...
-    <Compile Include="Adapters/SessionState/Serialization/SessionValues.cs" />
-    <Compile Include="Adapters/SessionState/Serialization/SerializedSessionState.cs" />
-    <Compile Include="Adapters/SessionState/Serialization/SessionSerializer.Shared.cs" />
+    <Compile Include="@(SharedFiles)" />
 
     <Reference Include="System.Web" />
-    <Reference Include="System.Configuration" />
-
-    <PackageReference Include="System.Text.Json" Version="6.0.0" />
-    <PackageReference Include="System.Threading.Channels" Version="6.0.0" />
```
Tratcher, line 58: The main goal was to remove these dependencies? What problems did we expect these to cause people?

Tratcher (approved):

src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/Microsoft.AspNetCore.SystemWebAdapters.csproj
twsouthwick: That the general HttpContext adapter usage could be done without a bunch of new dependencies. People often complain when their packages.config ends up with a ton of new packages (which is what this will do to a .NET 4.7.2 app). Since they may not need the session stuff, this can be a pay-for-play thing.

PR #497, opened 2022-04-20
Add support for IHttpHandler and friends

twsouthwick:


davidfowl:
I thought we agreed we weren't doing these.

twsouthwick:
It wasn't on the list, but seemed like pretty low hanging fruit. I'll close it and if we hear feedback about including it, I can bring it back

PR #500, opened 2022-04-20
Add instrumentation to adapters

twsouthwick:
This enables the ability to track how much effect the adapters are having on the execution of endpoints. This will also allow us to identify unnecessary work that is being done (i.e. session is turned on, but not being used), as well as how much reliance an endpoint has on the adapters.

### DIFF
--- a/src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
+++ b/src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
@@ -2,9 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Security.Claims;
 using System.Security.Principal;
+using System.Threading.Tasks;
 using System.Web.Caching;
 using System.Web.SessionState;
 using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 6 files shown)


twsouthwick:
An example of what the traces look like with Zipkin:

image

src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
```diff
@@ -28,15 +36,69 @@ public class HttpContext : IServiceProvider
...
+#endif
     }
 
-    public HttpRequest Request => _request ??= new(_context.Request);
+    public HttpRequest Request
+    {
+        get
+        {
+#if NET6_0_OR_GREATER
+            _activity?.AddEvent(new ActivityEvent(nameof(Request)));
```
Tratcher, line 64: This should only happen once, not every time the property is accessed. The same goes for the other properties.

src/SystemWebAdapters/src/Microsoft.AspNetCore.SystemWebAdapters/HttpContext.cs
```diff
@@ -46,7 +108,16 @@ public IPrincipal User
...
     }
 
-    public HttpSessionState? Session => _context.Features.Get<HttpSessionState>();
+    public HttpSessionState? Session
+    {
+        get
+        {
+#if NET6_0_OR_GREATER
+            _activity?.AddEvent(new ActivityEvent(nameof(Session)));
+#endif
```
Tratcher, line 117: This isn't lazily creating anything like the other properties. What cost art you tracking? The lookup should be almost free.

twsouthwick: The events can be used to understand the adapter usage for a given endpoint. 

One example is that using this telemetry we could identify that an endpoint is not using any adapter stuff off of HttpContext. That endpoint could then ensure it's not using any of the Middleware.

Another example would be to identify which endpoints use HttpRequest to bucket work to remove the adapter usage.

I see value in having an event for each access to get some idea of the amount of usage the adapter is getting as a way for users to prioritize moving endpoints off of the adapters.

PR #517, opened 2022-06-17
QuickGrid experimental package

SteveSandersonMS:
Initial implementation for https://github.com/dotnet/aspnetcore/issues/37908. This has evolved from the "grid" sample in our early rendering benchmark work, via the BlazeOrbital grid, into the current shape which is hopefully a lot more flexible while retaining the original performance.

The key techniques that QuickGrid uses to maximize performance are avoiding having child components for each cell, not having event handlers for each cell, and making careful use of `@key` and minimizing allocations. It should comfortably render tables around 10x100 with plain cell contents effortlessly. For much larger tables (or ones containing expensive child components), it makes it easy to enable UI virtualization, and can then comfortably scale up to 100,000 rows or whatever is the maximum supported by the browser.

### Scope

This is **not** intended to be a replacement for full-featured commercial grid products (such as those from component vendors). See the included README for details about what this is meant to achieve and what is the scope of this experimental package.

### Still to do

There's still a chunk of work remaining in setting up a demo site and a wider range of examples, along with installation and usage steps. In the future, we may choose to add features. There are a few basic ones (editing, column grouping) that I think are worth considering.

However, as a design goal we should **not** aim to include every possible grid feature. People who need the most sophisticated grids can continue to use those from component vendors. Hopefully grids from component vendors can take advantage of some of the architectural patterns present in QuickGrid to improve their performance and their degree of being idiomatic for Blazor and .NET generally.

### Review notes

Although this PR contains > 300 files, more than 200 of them are just country flag SVGs for the test data.

Nonetheless it's still a lot of code. I don't necessarily expect this to be reviewed with the same level of obsessive detail as we'd do for production-supported code. We can continue to change this (including in breaking ways) for quite a long time, maybe forever.


SteveSandersonMS:
**Blazor team:** BTW I might not wait for formal sign-off on this before proceeding, since it's an experimental labs package. But please do review if you're interested!

PR #534, opened 2022-07-21
Implement code to generate route handler code from a JSON file

anhthidao:
Write code to parse a JSON schema to an OpenApi Document and generate the route handler code from the information written in the JSON file.


src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
+namespace CodeGenerator
+{
+    using Microsoft.OpenApi.Models;
```
captainsafia, line 3: Using statements should be at the top of the file.

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
+namespace CodeGenerator
```
captainsafia, line 1: You can leverage top-level namespaces here so you don't have to nest everything under the curly brace.

Also, this file should be named `CodeGenerator` or something like that.

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
...
+{
+    using Microsoft.OpenApi.Models;
+    using Microsoft.OpenApi.Readers;
+    public class App
+    {
+        public static void Main(string[] args)
+        {
+            var document = new App().ReadJson("");
```
captainsafia, line 9: It's strange to pass `""` as a parameter here. Instead, you can make args nullable in the `ReadJson` method and not pass anything in here.

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
...
+
+                case "float":
+                    return "0.0f";
+
+                case "double":
+                    return "0.0d";
+            }
+            return "";
+        }
+        private OpenApiDocument ReadJson(string args)
```
captainsafia, line 152: `ReadJson` can be made into a static method ([reference](https://stackoverflow.com/a/4124118)). Static methods are not associated with an instance of a class so you can use them without having to call `new App()` above.

You can see static methods used heavily in the `OpenApiGenerator` class over in the aspnetcore repo.

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
...
+            return "";
+        }
+        private string GetPrimitiveValue(OpenApiSchema? schema)
+        {
+            var type = schema?.Type;
+
+            switch (type)
+            {
+                case "string":
+                    return "\"\"";
```
captainsafia, line 136:
```suggestion
                    return "string.Empty";
```

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
...
+                case "array":
+                    return new App().GetDataTypeKeyword(schema?.Items) + "[]";
+            }
+            return "";
+        }
+        private string GetPrimitiveValue(OpenApiSchema? schema)
+        {
+            var type = schema?.Type;
+
+            switch (type)
```
captainsafia, line 133: You can use switch expressions instead of switch statements here. See [this](https://csharp.christiannagel.com/2019/08/14/moving-from-the-switch-statement-to-switch-expressions-c-8/) for more info on the difference.

src/OpenApi/OutputFile/Program.cs
```diff
@@ -0,0 +1,6 @@
+var builder = WebApplication.CreateBuilder(args);
+var app = builder.Build();
+
+app.MapGet("/students", () => "");
+
+app.MapPost("/students/{id}", () => new string[][] {});
```
captainsafia, line 6: This looks a little off. It's returning an array of an array of strings (aka a matrix of strings) effectively. Correct?

Side note: we probably want to avoid commiting the generated files in.

src/OpenApi/CodeGenerator/RuntimeTextTemplate2.tt
```diff
@@ -0,0 +1,25 @@
...
+#>
+var builder = WebApplication.CreateBuilder(args);
+var app = builder.Build();
+<#
+}
+#>
+
+app.<#=method #>("<#=path #>", () => <#=returnValue #>);
+<#+
+    public string path { get; set; }
```
captainsafia, line 20: Nit: Properties are typically title-cased in C#, not camel-cased.

src/OpenApi/CodeGenerator/RuntimeTextTemplate2.tt
```diff
@@ -0,0 +1,25 @@
...
+    public string method { get; set; }
+    public string apiResponse { get; set; }
+    public bool shouldCreateWebApp { get; set; }
```
captainsafia, line 23: Why do we need this boolean? Feels like it'll always be true if we're generating an API...

src/OpenApi/CodeGenerator/RuntimeTextTemplate2.tt
```diff
@@ -0,0 +1,25 @@
...
+var app = builder.Build();
+<#
+}
+#>
+
+app.<#=method #>("<#=path #>", () => <#=returnValue #>);
+<#+
+    public string path { get; set; }
+    public string method { get; set; }
+    public string apiResponse { get; set; }
```
captainsafia, line 22: What is `apiResponse` used for?

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
...
+            String pageContent;
+            int countPaths = 0;
+            bool shouldCreateWebApp = true;
+
+            foreach (var path in paths)
+            {
+                var operations = path.Value.Operations;
+                foreach (var operation in operations)
+                {
+                    if (countPaths > 0)
```
captainsafia, line 24: It's probably better to check the length of `document.Paths` before even entering this loop. If there are no paths, then we should skip code generation and print a message to the user notifying them that no paths were found.

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
...
+                    {
+                        shouldCreateWebApp = false;
+                    }
+
+                    var method = operation.Key.ToString().ToLower();
+                    var response = operation.Value.Responses.FirstOrDefault().Value;
```
captainsafia, line 30: `FirstOrDefault` will return null if nothing is found. You probably want to check if the value is defined using `?` just in case.

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
...
+                    var schema = response.Content.Values.FirstOrDefault()?.Schema;
+
+                    var parameters = operation.Value.Parameters;
+                    string parametersList = "";
+                    int countParam = 1;
+
+                    foreach (var parameter in parameters)
```
captainsafia, line 37: Since you are incrementing `countParam` inside the body of this loop, it might be better to just use `for` instead of `foreach`.

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
...
+                    page = new RuntimeTextTemplate2
+                    {
+                        path = path.Key.ToString(),
+                        method = new App().GetHttpMethod(method),
+                        shouldCreateWebApp = shouldCreateWebApp,
+                        returnValue = returnValue,
+                        parametersList = parametersList
+                    };
+                    pageContent = page.TransformText();
+                    File.AppendAllText("C:\\Users\\AnhThiDao\\AspLabs\\src\\OpenApi\\OutputFile\\Program.cs", pageContent);
```
captainsafia, line 69: Avoid hard-coding the path here. I'd recommend adding `output-path` as an argument to the command line tool and reading the output path from that.

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
...
+                case "post":
+                    return "MapPost";
+
+                case "put":
+                    return "MapPut";
+
+                case "delete":
+                    return "MapDelete";
+
+                default:
```
captainsafia, line 99: Doesn't this mean we'll end up generating something like `app.(() => )` if we find an unexpected method name? It might be helpful to audit for this before generating a path.

src/OpenApi/CodeGenerator/Class1.cs
```diff
@@ -0,0 +1,212 @@
...
+                        parametersList = parametersList
+                    };
+                    pageContent = page.TransformText();
+                    File.AppendAllText("C:\\Users\\AnhThiDao\\AspLabs\\src\\OpenApi\\OutputFile\\Program.cs", pageContent);
+
+                    countPaths++;
+                }
+            }
+
+            //    //RuntimeTextTemplate2 page = new RuntimeTextTemplate2
```
captainsafia, line 75: Remove commented out code?

src/OpenApi/CodeGenerator/Class1.cs
anhthidao: this method is supposed to take in `args` as argument, but since we are testing and running everything in VS, we put a `""` there so that the method use the hard-coded path. we will change this in the final commit.

src/OpenApi/OutputFile/Program.cs
anhthidao: I confused the Java syntax with the C# syntax for multidimensional arrays. This will be fixed before the next commit. 

And I don't know why the generated file is also pushed. I didn't add them, but I guess it's auto-added.

src/OpenApi/CodeGenerator/RuntimeTextTemplate2.tt
anhthidao: this boolean is to make sure that the 
`var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();`
is only generated once in the output file

src/OpenApi/CodeGenerator/RuntimeTextTemplate2.tt
anhthidao: I forgot to remove it before commiting. I'll remove it now.

src/OpenApi/CodeGenerator/Class1.cs
anhthidao: yes the code would be written that way in the final commit. this is just for the purpose of testing in VS.

src/OpenApi/CodeGenerator/RuntimeTextTemplate2.tt
captainsafia: Yeah, I posted [this comment](https://github.com/aspnet/AspLabs/pull/534#discussion_r930179006) as well. We should always be generating the builder logic once.

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,190 @@
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Readers;
+
+namespace CodeGenerator;
+
+public class App
+{
+    public static void Main(string[] args)
+    {
+        if (args.Length == 0)
```
captainsafia, line 10: You probably want to check that the length is 2 to ensure that both the input and output file path are provided.

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,190 @@
...
+        var document = ReadJson(args[0]);
+        var paths = document.Paths;
+
+        if (paths is null || paths.Count == 0)
+        {
+            Console.WriteLine("No path was found in the schema!");
+            return;
+        }
+
+        RuntimeTextTemplate2 page;
```
captainsafia, line 24: Nit: Can we rename this to something clearer?

```suggestion
        MinimalApiTemplate page;
```

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,190 @@
...
+            if (operations is null || operations.Count == 0)
+            {
+                Console.WriteLine("No path was found!");
+                Environment.Exit(0);
+            }
+            foreach (var operation in operations)
+            {
+                if (countPaths > 0)
+                {
+                    shouldCreateWebApp = false;
```
captainsafia, line 42: This check doesn't seem meaningful. We'll also be creating the web app if paths are found. If there's not paths in the original OpenApi schema, then we should skip generation all together.

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,190 @@
...
+                }
+
+                var method = operation.Key.ToString().ToLower();
+                method = GetHttpMethod(method);
+
+                if (method == String.Empty)
+                {
+                    Console.WriteLine("Invalid operation found!");
```
captainsafia, line 50:
```suggestion
                    Console.WriteLine($"Unsupported HTTP method found: '{operation.Key}'");
```

src/OpenApi/CodeGenerator/CodeGenerator.cs
anhthidao: this check is to make sure that we only create the builder once. if we remove this check, there will be multiple `var builder = WebApplication.CreateBuilder(args);` statements printed to the output file

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,190 @@
...
+        "integer" => "0",
+        "boolean" => "false",
+        "float" => "0.0f",
+        "double" => "0.0d",
+        _ => String.Empty,
+    };
+    private static OpenApiDocument? ReadJson(string args)
+    {
+        //var inputPath = "C:\\Users\\AnhThiDao\\openapi.json";
+        var inputPath = "C:\\Users\\Anh Thi Dao\\Downloads\\petstore.json";
```
captainsafia, line 141: Remove hard-coded values.

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,190 @@
...
+            string path = Path.GetFullPath(args);
+            Stream stream = File.OpenRead(path);
+            OpenApiDocument newDocument = reader.Read(stream, out diagnostic);
+            return newDocument;
+        }
+        catch (FileNotFoundException e)
+        {
+            Console.WriteLine("Check to make sure you entered a correct file path because the file was not found.");
+            Console.Error.WriteLine(e.Message);
+            Environment.Exit(0);
```
captainsafia, line 163: `0` exit code is typically used for success. Failures should return a non-zero code. Typically, you would use a unique number for each failure case

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,190 @@
...
+            return null;
+        }
+        catch (Exception e)
+        {
+            Console.WriteLine("Check the file path you entered for errors.");
+            Console.Error.WriteLine(e.Message);
```
captainsafia, line 169: Let's use `Console.Error.WriteLine` for all error reporting so it is sent to stderr.

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,195 @@
...
+            Console.Error.WriteLine("Please enter two arguments: an input file path and an output file path.");
+            Environment.Exit(1); // Code 1 is for problems with passed in file paths
+        }
+
+        var document = ReadJson(args[0]);
+        var paths = document?.Paths;
+
+        if (paths is null || paths.Count == 0)
+        {
+            Console.Error.WriteLine("No path was found in the schema.");
```
captainsafia, line 22:
```suggestion
            Console.Error.WriteLine("No path were found in the schema.");
```

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,195 @@
...
+                if (method == String.Empty)
+                {
+                    Console.Error.WriteLine($"Unsupported HTTP method found: '{operation.Key}'");
+                    Environment.Exit(3);
+                }
+
+                fileProperties[pathString].Add(method, new Dictionary<string, string> { });
+
+                var parameters = operation.Value.Parameters;
+                string parametersList = "";
```
captainsafia, line 53:
```suggestion
                string parametersList = string.Empty;
```

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,195 @@
...
+            {
+                Console.WriteLine("Read File Successfully");
+            }
+            else
+            {
+                foreach (OpenApiError error in diagnostic.Errors)
+                {
+                    Console.WriteLine($"There was an error reading in the file: {error.Pointer}");
+                    Console.Error.WriteLine(error.Message);
+                    Environment.Exit(1);
```
ilmax, line 190: this exit at the first error, stopping the iteration

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,305 @@
...
+            }
+            var pathString = path.Key.ToString();
+            fileProperties.Add(pathString, new Dictionary<string, Dictionary<string, string>> { });
+
+            foreach (var operation in operations)
+            {
+                var method = operation.Key.ToString().ToLower();
+                method = GetHttpMethod(method);
+
+                if (method == String.Empty)
```
captainsafia, line 44:
```suggestion
                if (method == string.Empty)
```

Here and elsewhere.

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,305 @@
...
+        var page = new MinimalApiTemplate
+        {
+            FileProperties = fileProperties,
+            Schemas = schemaDict
+        };
+
+        var pageContent = page.TransformText();
+        File.WriteAllText(args[1], pageContent);
+    }
+    private static string GetSampleValue(IOpenApiAny example, OpenApiSchema? schema) => example switch
```
captainsafia, line 163: Nit: make sure there are newlines before method definitions.

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,305 @@
...
+                    if (response.Value.Content == null || response.Value.Content.Count == 0)
+                    {
+                        returnValue = "Default";
+                        fileProperties[pathString][method].Add(response.Key, returnValue);
+                        continue;
+                    }
+                    var content = response.Value.Content.First().Value;
+                    var schema = content.Schema;
+
+                    if (schema?.Type.ToLower() == "array")
```
captainsafia, line 96: Maybe instead of using `schema?` repeatedly here, we can check to see if a schema is defined above front and skip this logic altogether if it is not.

src/OpenApi/CodeGenerator/CodeGenerator.cs
```diff
@@ -0,0 +1,305 @@
...
+
+                foreach (var response in responses)
+                {
+                    string returnValue;
+
+                    // some responses doesn't have "content" property
+                    // so these would later return a default value
+                    if (response.Value.Content == null || response.Value.Content.Count == 0)
+                    {
+                        returnValue = "Default";
```
captainsafia, line 89:
```suggestion
                        returnValue = "default";
```

src/OpenApi/CodeGenerator/MinimalApiTemplate.tt
```diff
@@ -0,0 +1,132 @@
...
+
+<#+
+    public Dictionary<string, Dictionary<string, Dictionary<string, string>>> FileProperties { get; set; }
+    public string Path { get; set; }
+    public string Method { get; set; }
+    public string ReturnValue { get; set; }
+    public string ParametersList { get; set; }
+    public string StatusCode { get; set; }
+    public Dictionary<string, Dictionary<string, string>> Schemas { get; set; }
+    public string UserObject { get; set; }
```
captainsafia, line 127: `UserObject` is a slightly confusing name here. This is referring to a schema type that is a POCO, yes? It might be sufficient to call it `CustomObject` or something. `UserObject` implies it's related to identity/authentication which is not the case.

src/OpenApi/CodeGenerator/MinimalApiTemplate.tt
```diff
@@ -0,0 +1,132 @@
...
+    }
+    public <#=UserObject #>() {}
+}
+<#
+    }
+#>
+
+<#+
+    public Dictionary<string, Dictionary<string, Dictionary<string, string>>> FileProperties { get; set; }
+    public string Path { get; set; }
```
captainsafia, line 121: It looks like only FileProperties and Schemas are actually set from outside of the template. All the other things don't need to be public properties and can just be local variables in template code. Is there a reason public properties are used?

captainsafia (approved):

PR #547, opened 2022-09-01
Update public pool names

wtgodbe:


dougbu (approved):

wtgodbe:
@alexperovich do we need the same changes here as in https://github.com/aspnet/AspNetWebHooks/pull/154?

alexperovich:
Yea, we do.

wtgodbe:
Actually, wait, I don't see anywhere where we set an endpoint in a .yml file - is it something else?

alexperovich:
Not that change. I need to update the build to use the workaround so I need perms to change the webhooks.

wtgodbe:
I gave you admin rights

alexperovich:
Okay, it looks fixed now.

PR #550, opened 2022-10-20
explicitly identifying button type in paginator

ebekker:
As per the [HTML spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type), if a button's `type` is not explicitly specified, then it defaults to a `submit` type button.  This causes issues if the default QuickGrid Paginator is placed inside of an EditForm, as clicking any nav button triggers the EditContext's validation.


Nick-Stanton:
Linked to https://github.com/dotnet/aspnetcore/issues/45216

SteveSandersonMS:
Makes sense. Thanks!

Nick-Stanton (approved):

PR #551, opened 2022-10-26
Add CodeQL3000 run to AspLabs pipeline

wtgodbe:
Test build here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2030291&view=results. Runs weekly Monday mornings at 2


azure-pipelines.yml
```diff
@@ -57,9 +79,24 @@ stages:
...
             name: NetCore1ESPool-Internal
             demands: ImageOverride -equals Build.Server.Amd64.VS2017
+        ${{ if eq(variables.runCodeQL3000, 'true') }}:
+          # Component governance and SBOM creation are not needed here. Disable what Arcade would inject.
+          disableComponentGovernance: true
+          enableSbom: false
         variables:
         - name: _HelixBuildConfig
           value: $(_BuildConfig)
+        - ${{ if eq(variables.runCodeQL3000, 'true') }}:
```
dougbu, line 89: Current asplabs-ci builds run CG in public. Will also uselessly auto-inject CG in internal builds, only to be skipped because the Arcade task runs first.
```suggestion
        # Rely on task Arcade injects, not auto-injected build step.
        - skipComponentGovernanceDetection: true
        - ${{ if eq(variables.runCodeQL3000, 'true') }}:
```

wtgodbe: I'll commit this here since it's nice to have, but I haven't been actively looking for things like this in the other aspnet/* repos since it's unrelated to CodeQL

wtgodbe:
@dougbu any other concerns?

azure-pipelines.yml
```diff
@@ -92,120 +131,132 @@ stages:
           inputs:
             command: custom
             arguments: 'locals all -clear'
+        - ${{ if eq(variables.runCodeQL3000, 'true') }}:
+          - script: "echo ##vso[build.addbuildtag]CodeQL3000"
```
dougbu, line 135: Move this after the `...Init` task

azure-pipelines.yml
```diff
@@ -92,120 +131,132 @@ stages:
...
+          - task: PublishBuildArtifacts@1
+            displayName: Publish Packages
+            condition: and(eq(variables['system.pullrequest.isfork'], false), eq(variables['_BuildConfig'], 'Release'))
+            continueOnError: true
+            inputs:
+              artifactName: Packages_$(Agent.Os)_$(Agent.JobName)
+              parallel: true
+              pathtoPublish: '$(Build.SourcesDirectory)/artifacts/packages/$(_BuildConfig)'
+              publishLocation: Container
+          - task: PublishBuildArtifacts@1
```
dougbu, line 168: Suggest always publishing the logs. They're often very useful.

tsaoptions.json
```diff
@@ -0,0 +1,12 @@
+{
```
dougbu, line 1: This file won't be found. The tasks expect to find it in `{Source Root}/.config/` unless you set `Codeql.TSAOptionsPath`.

azure-pipelines.yml
wtgodbe: What difference does it make?

dougbu: As-is, the `condition` will never be `true`; `CodeQL3000Init@0` adds the variable when it and the `...Final` task are going to do real work.

tsaoptions.json
wtgodbe: Fixed

azure-pipelines.yml
dougbu: My point was: Undo the indentation change you made for this `task` so that it runs in `CodeQL3000` builds too.

wtgodbe: Ah, fair enough, updated

dougbu (approved):

PR #572, opened 2023-01-04
Add .Net 7 Check to ClientAssets

Nick-Stanton:
While the effort continues to productionize Microsoft.AspNetCore.ClientAssets, there needs to be an updated check to make sure that customers upgrading to .Net 7 don't break.

CC @mkArtakMSFT 

Fixes dotnet/aspnetcore#45849

### DIFF
--- a/src/ClientAssets/Microsoft.AspNetCore.ClientAssets/build/netstandard2.0/Microsoft.AspNetCore.ClientAssets.targets
+++ b/src/ClientAssets/Microsoft.AspNetCore.ClientAssets/build/netstandard2.0/Microsoft.AspNetCore.ClientAssets.targets
@@ -47,7 +47,7 @@
       <FileWrites Include="$(IntermediateOutputPath)clientassetsbuild.complete.txt" />
 
       <Content
-        Condition="'$(TargetFramework)' != 'net6.0'"
+        Condition="'$(TargetFramework)' != 'net6.0' Or '$(TargetFramework)' != 'net7.0'"
         Include="@(_ClientAssetsBuildOutput)"
         Link="wwwroot\%(_ClientAssetsBuildOutput.RecursiveDir)\%(_ClientAssetsBuildOutput.FileName)%(_ClientAssetsBuildOutput.Extension)" />
 
@@ -58,7 +58,7 @@
       the ability to consume them from project references. As a workaround, we are using the Task directly on this package, to make sure this scenario works. In a future
       release we will address this by aligning DiscoverStaticWebAssets behavior with DefineStaticWebAssets to follow the same heuristics for determining the content root.
     -->
-    <DefineStaticWebAssets Condition="'$(TargetFramework)' == 'net6.0'"
+    <DefineStaticWebAssets Condition="'$(TargetFramework)' == 'net6.0' Or '$(TargetFramework)' == 'net7.0'"
       CandidateAssets="@(_ClientAssetsBuildOutput)"
       SourceId="$(PackageId)"
       SourceType="Computed"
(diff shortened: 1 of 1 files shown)


src/ClientAssets/Microsoft.AspNetCore.ClientAssets/build/netstandard2.0/Microsoft.AspNetCore.ClientAssets.targets
```diff
@@ -58,7 +58,7 @@
       the ability to consume them from project references. As a workaround, we are using the Task directly on this package, to make sure this scenario works. In a future
       release we will address this by aligning DiscoverStaticWebAssets behavior with DefineStaticWebAssets to follow the same heuristics for determining the content root.
     -->
-    <DefineStaticWebAssets Condition="'$(TargetFramework)' == 'net6.0'"
+    <DefineStaticWebAssets Condition="'$(TargetFramework)' &gt;= 'net6.0'"
```
javiercn, line 61: This does not work in MSBuild 😄.

You are also comparing strings, not numbers. Check how the Razor SDK does it https://github.com/dotnet/sdk/blob/main/src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets#L46-L91

Nick-Stanton: Lame! I misread how version parsing worked in MSBuild. Incoming commit to fix this, but does there need to be compatibility for .NET Standard?

javiercn: Nope, only net6.0+

Nick-Stanton:
Blocked pending the resolution of an email thread about whether this PR would be good opportunity to add a project capability.

src/ClientAssets/Microsoft.AspNetCore.ClientAssets/build/netstandard2.0/Microsoft.AspNetCore.ClientAssets.targets
```diff
@@ -18,6 +18,7 @@
 
   <ItemGroup>
     <ClientAssetsInputs Include="$(ClientAssetsDirectory)**" Exclude="$(DefaultItemExcludes)" />
+    <ProjectCapability Include="ClientAssetsAlpha"/>
```
mkArtakMSFT, line 21: My preference would be to not include `Alpha` suffix. I wonder if it'll be possible to differentiate telemetry based on the package version itself (@DamianEdwards do you know?)

DamianEdwards: No we don't get good signal on package versions.

javiercn: I want a signal for these versions so once we decide to change it, if it becomes a package, we can differentiate usages as @DamianEdwards says.

javiercn (approved):

lscorcia:
Can we get an updated NuGet for this? Even prerelease would be fine. It seems that the version at https://www.nuget.org/packages/Microsoft.AspNetCore.ClientAssets/ is not the latest and this prevents us to update to .net 7.0.

javiercn:
@mkArtakMSFT did you publish the package for this?

austinmcdaniel:
Any updates on getting another nuget release for this change -- even an alpha?
@javiercn @mkArtakMSFT

javiercn:
@mkArtakMSFT can you push an updated version of the package?

PR #578, opened 2023-02-17
explicitly identifying button type in columnbase

Lorcan-exida:
Linked to https://github.com/dotnet/aspnetcore/issues/45216

Making Header Column buttons explicitly of `type=button`

This will ensure that when the default QuickGrid Header has `ColumnOptions` and/or Sorting enabled inside an `EditForm` it will no longer trigger the `EditContext` validation.

This fix is a continuation to #550


Nick-Stanton:
Hi @Lorcan-exida, thank you for your contribution. QuickGrid has now been productized in dotnet/aspnetcore#46317 and will no longer receive changes here. When convenient, please open a new PR against the version in the ASP.NET Core repository.

PR #579, opened 2023-02-17
Update GridSort.cs

drualcman:
Simplify ToPropertyList because no need the if. Can send directly the parameter to BuildProperyList


drualcman:
Have a mistake becasue copy and pase, already create a new pull request fixing my mistake. Sorry.

PR #580, opened 2023-02-17
Update GridSort.cs

drualcman:
Simplify ToPropertyList because no need the if to call BuildPropertyList. Fixed previous mistake in copy and paste action no set correctly the parameter.


drualcman:
Because is Nulable IF it's required to avoid warning  warning CS0414: The field 'GridSort<TGridItem>._cachedPropertyListDescending' so discard this pull request, sorry my mistake i didn't read correctly the code.

PR #581, opened 2023-02-24
Recreate package-lock.json files

dougbu:
- bump all resolved dependencies to latest compatible versions
  - re-resolve everything but don't change any package.json files
- bump `json5`, `loader-utils`, and `minimatch` in particular
  - all the subject of recent Dependabot PRs
- also add `always-auth=true` to .npmrc files
  - required to ingest packages into `dotnet-public-npm`

Note: Unable to use just one .npmrc file here
- appears `yarn` looks up the tree by default but `npm` does not

### DIFF
--- a/src/ClientAssets/samples/BlazorWasmAppWithJsAssets/assets/.npmrc
+++ b/src/ClientAssets/samples/BlazorWasmAppWithJsAssets/assets/.npmrc
@@ -1,1 +1,2 @@
 registry=https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/
+always-auth=true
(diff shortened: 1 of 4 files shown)


mkArtakMSFT:
Thanks @dougbu.
One of the major changes I notice here is that the source for all the packages have switched from `registry.npmjs.org` to `pkgs.dev.azure.com/dnceng/public`. I assume this is intentional? Just wanted to make sure that's the case.

dougbu:
> Thanks @dougbu. One of the major changes I notice here is that the source for all the packages have switched from `registry.npmjs.org` to `pkgs.dev.azure.com/dnceng/public`. I assume this is intentional? Just wanted to make sure that's the case.

Yes, that was intentional. The .npmrc files were _meant_ to ensure we resolved our `npm` packages from `dotnet-public-npm`. Without this change, that wasn't put into practice.

wtgodbe (approved):

PR #584, opened 2023-04-10
Fix component governance alerts

MackinnonBuck:
Resolves CVE-2021-1723


wtgodbe (approved):
If it works - though this one could also just be dismissed since it's only in samples/tests

MackinnonBuck:
Actually, https://github.com/aspnet/AspLabs/pull/583 already effectively makes this change. Let's merge that instead.

PR #591, opened 2023-08-28
Delete Proxy, WebHooks from repo

wtgodbe:
These projects are no longer in use, and are just accumulating Component Governance debt.


mkArtakMSFT (approved):

Tratcher (approved):

astro-nahuel:
I arrived to this repo following the link in this page: https://github.com/aspnet/aspnetwebhooks?tab=readme-ov-file and got a 404 inside this repo, so I then searched for "webhooks" in the PRs and ended up here.

Are webhooks no longer a first-class citizen in ASP.NET, or has this code simply moved somewhere else, or does the feature have a different name now? I'm ok with doing it on my own from scratch, I just want to be sure the feature is being removed completely and I'm not reinventing the wheel in vain.

wtgodbe:
Only the .NET Framework version of webhooks is still in support - we currently do not plan to bring the feature into core .NET: https://github.com/dotnet/aspnetcore/issues/48438

PR #605, opened 2024-08-19
Enhance XML comment extraction for Swagger docs

paulomorgado:
Update GrpcXmlCommentsDocumentFilter.cs to prioritize "value" XML tags for Swagger descriptions, falling back to "summary" tags if "value" is not present. Introduce `ValueTag` constant and modernize null checks with `is not null` syntax. This improves the accuracy and specificity of the generated Swagger documentation.

### DIFF
--- a/.config/tsaoptions.json
+++ b/.config/tsaoptions.json
@@ -1,0 +1,12 @@
+{
+  "areaPath": "DevDiv\\ASP.NET Core\\Policy Violations",
+  "codebaseName": "AspLabs",
+  "instanceUrl": "https://devdiv.visualstudio.com/",
+  "iterationPath": "DevDiv",
+  "notificationAliases": [
+    "aspnetcore-build@microsoft.com"
+  ],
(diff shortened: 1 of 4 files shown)


paulomorgado:
@captainsafia, this is what I was talking about.

Couldn't figure out how to test this. Tests seem to only use protobuf.

captainsafia:
@paulomorgado That code that you've modified here is for gRPC's OpenAPI + XML implementation. This is a separate experiment that was resolved a few releases ago and now lives in the ASP.NET Core repo. The XML implementation is in  [this directory](https://github.com/aspnet/AspLabs/tree/safia/openapi-xml/src/OpenApiXmlCommentGenerator).

For this specific case, I'd recommend filing an issue on the dotnet/aspnetcore repo specifically around how the gRPC implementation handles `<value>`.

For the OpenAPI XML support (seperate from gRPC), I'd recommend waiting until things are merged to main before opening PRs.

PR #609, opened 2024-10-17
Merge OpenApiXmlComment generator to main

captainsafia:
This has been sitting in a branch for a while but merging it to main so we can churn out a NuGet package and start to iterate on this prototype more seriously.


global.json
```diff
@@ -1,15 +1,18 @@
...
+        "9.0.0-rc.2.24473.5",
+        "6.0.0"
       ],
       "aspnetcore": [
-        "6.0.3"
+        "9.0.0-rc.2.24474.3",
+        "6.0.0",
+        "3.1.23"
       ],
       "aspnetcore/x64": [
```
wtgodbe, line 17: Might as well remove this whole block while you're here (x64 & x86)

global.json
```diff
@@ -1,15 +1,18 @@
...
       "dotnet": [
-        "6.0.3"
+        "9.0.0-rc.2.24473.5",
+        "6.0.0"
       ],
       "aspnetcore": [
-        "6.0.3"
+        "9.0.0-rc.2.24474.3",
+        "6.0.0",
+        "3.1.23"
```
wtgodbe, line 15: Remove

wtgodbe (approved):
Didn't look at the content of the addition, but the changes generally look good

global.json
```diff
@@ -13,12 +13,6 @@
         "9.0.0-rc.2.24474.3",
         "6.0.0",
         "3.1.23"
-      ],
```
wtgodbe, line 13: You might actually need to keep this, I think the format was different in 3.1

global.json
captainsafia: Per offline conversation, there are still projects that target `netcoreapp3.1` in this repo. We can probably think about deleting them separately.

global.json
captainsafia: I suppose CI will let us know soon enough?

PR #612, opened 2025-01-06
Fix generic type handling

mentallabyrinth:
Love the direction `Microsoft.AspNetCore.OpenApi.SourceGenerators` is going, but there's a bug when extracting comments from types that support generics. The issue is in [https://github.com/aspnet/AspLabs/blob/main/src/OpenApiXmlCommentGenerator/gen/XmlCommentGenerator.Parser.cs](https://github.com/aspnet/AspLabs/blob/main/src/OpenApiXmlCommentGenerator/gen/XmlCommentGenerator.Parser.cs), specifically in these lines of code:

```csharp
var typeInfo = type.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat);
var typeInfo = property.ContainingType.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat);
```

To fix this, I've added a new method `ToNormalizedDisplayString` to `INamedTypeSymbolExtensions.cs`. The method checks if the type is generic - if so, it constructs an unbounded generic type (which works well for the comment caching strategy) and uses that for the display name. Otherwise, it uses the general type:

```csharp
public static string ToNormalizedDisplayString(this INamedTypeSymbol symbol, SymbolDisplayFormat? format = null)  
{  
    if (symbol.IsGenericType)  
    {        var genericType = symbol.ConstructUnboundGenericType();  
        return genericType.ToDisplayString(format);  
    }  
    return symbol.ToDisplayString(format);  
}
```

Looking forward to the general release. Thanks for your efforts!

### DIFF
--- a/src/OpenApiXmlCommentGenerator/gen/Helpers/INamedTypeSymbolExtensions.cs
+++ b/src/OpenApiXmlCommentGenerator/gen/Helpers/INamedTypeSymbolExtensions.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using Microsoft.CodeAnalysis;
 
 public static class INamedTypeSymbolExtensions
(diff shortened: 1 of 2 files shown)


mentallabyrinth:
@dotnet-policy-service agree

PR #615, opened 2025-01-20
Update gRPC HTTP API README.md

JamesNK:


BrennanConroy (approved):
Any reason not to just delete this directory?

JamesNK:
There are some social media posts that link to it. Leaving it around with a link to the real source code seems a simple short term solution.

Eventually it could be deleted.

